PR #24, opened 2026-04-19
feat: show cursor block in exec terminal display

tsacha:
Really simple commit to show cursor in the exec view:

image


janosmiko:
Thanks @tsacha 🙏

I added these changes in e0f4d43, also added a check to only show the cursor, when the cursor is visible.

PR #25, opened 2026-04-19
feat: send arrow key sequences in DECCKM application in exec view

tsacha:
Hey,

For now, in the exec view some applications like htop doesn't work properly with arrow keys.
This PR ensures arrow keys send the correct sequences when the terminal is in Application Cursor Mode (DECCKM).

Tell me what you think!

(used AI for this one)


janosmiko:
Thank you for your contribution @tsacha !

PR #30, opened 2026-04-20
fix: key namespace completion cache by context

### LINKED ISSUE #29: Command mode: namespace cache is global, leaks across tabs/contexts
## Problem

The command-bar namespace completion cache (`Model.cachedNamespaces`) is a single flat `[]string` with no context key. It's populated once on the first `:` keypress and then reused forever — which breaks two common workflows:

### Use case 1: two tabs, two clusters

1. Tab 1 on cluster **A** (context `ctx-a`) — press `:`, the app fetches A's namespaces into `cachedNamespaces`.
2. Open a new tab, drill into cluster **B** (context `ctx-b`) — tab state correctly tracks `nav.Context = ctx-b`.
3. Press `:` in tab 2 and type `:po <substring>`.

Expected: dropdown shows namespaces from **cluster B**.
Actual: dropdown shows namespaces from **cluster A** (the old cache). The empty-check at `internal/app/update_keys.go:767` is `len(m.cachedNamespaces) == 0` — because the cache is non-empty (filled from A), no refetch is triggered for B.

This is particularly confusing because `nav.Context`, `discoveredResources`, and `commandBarNameCache` are all per-context-correct — namespaces are the odd one out.

### Use case 2: single tab, `:ctx` switch

1. Open the app on `ctx-a`, press `:` — cache fills with A's namespaces.
2. Run `:ctx ctx-b` — `nav.Context` updates, but `cachedNamespaces` is not invalidated.
3. Type `:po <substring>` — dropdown still filters over `ctx-a`'s namespaces.

Same root cause.

## Root cause

- `Model.cachedNamespaces` is `[]string` — flat, no context key (`internal/app/app.go:751`).
- `handleKeyCommandBar` only triggers `loadNamespaces` when the list is empty (`internal/app/update_keys.go:767`). Once populated for any context, it never refreshes.
- Context-switch sites (`commandbar_execute.go:153` `:ctx`, `update_navigation.go:240` UI select, `update_bookmarks.go:423, 550` session restore) don't invalidate the cache either.
- Per-tab `saveCurrentTab` / `loadTab` (`internal/app/tabs.go:650-845`) doesn't copy `cachedNamespaces` — it's a shared global.

## Proposed fix

Key the cache by context name, matching the pattern already used for `discoveredResources` and `commandBarNameCache`:

```go
cachedNamespaces map[string][]string   // was []string
```

- `namespaceNames()` returns `m.cachedNamespaces[m.nav.Context]`.
- `namespacesLoadedMsg` gets a `context` field so the async reply is stored under the context the fetch was issued for (guards against a race where the tab moves during the fetch).
- The empty-check becomes `len(m.cachedNamespaces[kctx]) == 0` — so each context's first `:` open triggers its own fetch.
- No clearing on context switch needed; previously-loaded contexts stay cached and switching back is instant.
- No per-tab state copying needed; the map is cross-tab but per-context-correct.

## Repro environment

Any setup with at least two kube contexts and at least one namespace in context B whose name doesn't prefix-match any namespace in context A (so the stale-cache hit is unambiguous).

AndyHoang:
Closes #29

## Summary

The command-bar namespace cache (`Model.cachedNamespaces`) is a flat `[]string` populated once on the first `:` keypress. Once filled, the empty-check at `update_keys.go:767` (`len(m.cachedNamespaces) == 0`) never triggers again — so every tab and every `:ctx` switch reads the first context's namespaces, even after `nav.Context` has moved.

This PR re-keys the cache by context name so each tab / context gets its own entry.

### Two tabs, two clusters

1. Tab 1 on `ctx-a` → press `:` → fetches A's namespaces.
2. Open tab 2 on `ctx-b` → press `:` → type `:po <substring>`.
3. **Before:** dropdown shows A's namespaces (stale cache).
4. **After:** dropdown shows B's namespaces. Switching back to tab 1 reuses A's cached entry (no re-fetch).

### Single tab, `:ctx` switch

1. Cache fills for `ctx-a`.
2. Run `:ctx ctx-b` → `nav.Context` updates but the flat cache was never invalidated.
3. **Before:** `:po <substring>` filters over A's namespaces.
4. **After:** B's entry is missing, so `:` triggers a fetch for B, stored under its own key.

## Implementation

- `Model.cachedNamespaces` → `map[string][]string` (`app.go:751`).
- `namespacesLoadedMsg` gains a `context` field so async replies land under the context the fetch was issued for (guards against the tab moving during the fetch) (`messages.go:60`, `commands_load.go:242`).
- `updateNamespacesLoaded` stores under `msg.context`; lazy-inits the map (`update.go:837-846`).
- `namespaceNames()` returns `m.cachedNamespaces[m.nav.Context]`, falling back to `m.client.CurrentContext()` if nav hasn't moved off the root yet (`commandbar_complete.go:645-655`).
- `handleKeyCommandBar` empty-check becomes per-context (`update_keys.go:760-771`).

Matches the per-context keying already used by `discoveredResources` and `commandBarNameCache` — no new pattern introduced.

No clearing on context switch needed; previously-loaded contexts stay cached. No per-tab `TabState` copying needed; the map is cross-tab but per-context-correct.

## Caveats / follow-ups (out of scope for this PR)

- No TTL or invalidation. Namespaces created/deleted in-cluster after the first fetch won't appear until the app restarts. If that matters, a simple `:ctx` switch could force a refresh, or we could add a TTL / watch hook — happy to do that in a follow-up PR.

## Test plan

- [x] `go test ./...` — all passing, including 3 new tests:
  - `TestNamespaceNames_ScopedByContext` — per-context read is isolated
  - `TestCompleteBuiltin_NamespaceArg_PerContext` — no leak across contexts in the actual completion path
  - `TestUpdateNamespacesLoaded_StoresUnderMessageContext` — tab-moved-during-fetch race doesn't misroute the result
- [x] `go build -o lfk .` — clean
- [x] Reviewer: reproduce the two-tab case above on a real cluster and confirm B's namespaces appear, then tab back and confirm A's reappear without a re-fetch

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -69,6 +69,8 @@ Repeated completed tasks collapse with a `×N` suffix.
 
 Autocomplete offers subcommands, flags, resource types, and namespaces. Typing `:k` or `:kubectl` alone surfaces the subcommand list. Value positions (namespace, resource name, output format) accept fuzzy matches — exact > prefix > substring > subsequence — while command names themselves stay on prefix.
 
+Namespace suggestions come from a per-context cache warmed when the context is opened and refreshed on a 60s TTL. In-app mutations — `:k create ns`, `:k delete ns`, or a template apply — invalidate the cache immediately so new namespaces appear in completions without waiting for the TTL; changes made outside the TUI (CI, cloud console, `kubectl` in another shell) surface on the next refresh. The existing list stays visible during refreshes so completions never blank out.
+
 ## Resource jumps
 
 Type a resource name to navigate. Plural, singular, and kubectl abbreviations all work.
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -750,9 +750,11 @@ type Model struct {
 	// Cached namespace names for command bar autocompletion, keyed by
 	// context name. Each tab may have its own nav.Context, so keying by
 	// context keeps completions correct when switching tabs or running
-	// `:ctx` within a tab without re-fetching a context's namespaces more
-	// than once.
-	cachedNamespaces map[string][]string
+	// `:ctx` within a tab. Entries carry a fetchedAt timestamp so the
+	// command bar can refresh them after namespaceCacheTTL without
+	// refetching on every open (stale-while-revalidate: the old entry
+	// stays visible while the refresh runs).
+	cachedNamespaces map[string]namespaceCacheEntry
 
 	// Async resource name cache for cross-namespace kubectl completion.
 	// Key: "context/namespace/resource" -> list of resource names.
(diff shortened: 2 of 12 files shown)


janosmiko:
Thanks for the contribution @AndyHoang !

PR #99, opened 2026-04-30
feat(ui): dim explorer behind overlays via dim_overlay option

janosmiko:
## Summary
- Fade the explorer to a faint dark grey while any overlay is up so the dialog stands out and the bottom hint bar (kept un-faded) carries the keymap.
- New `ui.DimBackground` helper renders dim with a raw `\x1b[2;38;5;240m...\x1b[0m` (SGR faint + 256-gray) so the effect reaches terminals where lipgloss might otherwise strip styling. `keepLast = 1` preserves the hint bar verbatim; line count is preserved so `PlaceOverlay` alignment stays correct.
- Configurable via `dim_overlay` (default `true`) in `~/.config/lfk/config.yaml` for terminals where SGR faint looks awkward. Documented in `docs/config-reference.md` and the example config.
- Two deliberate skips: `overlayNone` (no-op renders) and `overlayColorscheme` (the picker exists to preview themes against the live explorer, so the comparison must stay vivid).

## Test plan
- [x] `go test -race ./...` is clean
- [x] In a real terminal: open Quit, Confirm, NetworkPolicy, CanI/Subject overlays — explorer fades, hint bar stays bright
- [x] Open colourscheme picker — explorer remains bright (theme previews must read correctly)
- [x] Set `dim_overlay: false` in config → no fading at all
- [x] Test on a low-colour profile (e.g. `TERM=screen`) to confirm raw SGR still applies


janosmiko:
Hi @mudejar, @Necrom4,

As you both opened a PR to update the quit overlay text, I made some changes. Shrank the quit window and added the y/n to the hint bar on the bottom.

On top of this, I implemented a functionality to dim the "background" when an overlay appears (quit, namespace selector, actions overlay, etc.).

However, I'm not really sure if I want to enable this by default.

This is how it looks like. What do you think?

CleanShot 2026-04-30 at 23 07 10@2x

CleanShot 2026-04-30 at 23 07 47@2x


I'm really curious about your opinions too @sp71, @shoce

Necrom4:
That's great, I think for such simple things there shouldn't be a need to configure it, just have it as a default. I'd recommend dimming less though, so that text is darker but not so "faded" in the background.

Also the content in both windows of your screenshots seems to be misaligned one character length to the left in relation to the borders.

janosmiko:
> I think for such simple things there shouldn't be a need to configure it

It's not a F*CK YEAH, LET'S DO THIS feature for me (I thought it will be, but it's not). And I'm not even sure if I want to enable it as a default. That's why I'm asking for your opinions.

shoce:
i think dimmed background is cool, it should be on by default, does not even need a config option. my config is 
```
transparent_background: true
no_color: true
icons: "none"
mouse: false
```
but if i would accidentally start lfk without these settings, it would make complete sense to see the dim background when a dialog is shown.

mudejar:
I agree that having it on by default is probably the right move. Also +1 for what @Necrom4 said because sometimes a programmer may want to take a quick look at some text in the background after the overlay appears because he forgot something, so it should still be legible even while dimmed.

sp71:
Dimming in general is fine but it changes background color (black & white) instead of just dimming theme color.  A bit jarring experience. It's not a feature I need but if others find it useful, sure.

**Inconsistent UX:**

Help (?) is a fullscreen mode, not an overlay. It replaces the explorer view entirely, so there's no background to dim. The dim_overlay setting only kicks in for popup overlays that float on top of the explorer.

**Flicker Issue:**

1. go to pod list view
2. j to another pod
3. `\`
4. the top header bar flickers (lfk > context > Pods)

Issue #154, opened 2026-05-05
[feat]: adopt vim subset for keybindings instead of custom variants


sp71:
## Problem / motivation

- lfk has invested heavily in vim-style bindings in the viewers (YAML/Log/Diff/Describe/Events): `hjkl`, `w`/`b`/`e`, `gg`/`G`, `0`/`^`/`$`, `/`+`n`/`N`, `v`/`V`/`C-v`, `C-d`/`C-u`/`C-f`/`C-b` with vim `'scroll'` semantics, marks, counts.
- Some current bindings diverge from vim: `y` alone yanks the current line (vim: operator that waits for a motion; line-yank is `yy`).
- As more bindings get added piecemeal, custom variants and vim-correct ones risk drifting further apart, which hurts muscle memory for the audience that already reaches for these keys.

## Proposed solution

Adopt a guiding principle: **when adding a binding that has an established vim/nvim meaning, match vim semantics; don't invent variants.**

Concrete bindings to add under this principle (viewers: YAML / Log / Diff / Describe / Events):

- `f<char>` / `F<char>` / `t<char>` / `T<char>` — find/till char on current line
- `*` / `#` — search word-under-cursor forward/backward (reuses existing `/` + `n`/`N`)
- `zz` / `zt` / `zb` — center / top / bottom the current line in the viewport
- `y` + motion (operator-pending) — `yw`, `yW`, `ye`, `yb`, `y$`, `y0`, `y^`, `yfX`, `ytX`
- `yiw` / `yi"` / `yi'` — yank inner word / inner quoted string (high value for grabbing values out of YAML)

Breaking change required to make operator-pending work cleanly — scoped to the viewers only (surfaces with a cursor moving through text):

- In viewers, `y` alone becomes an operator (waits for a motion), no timeout-fallback hybrid.
- In viewers, `yy` becomes the line-yank shortcut (`5yy` for 5 lines, matching `5y` today).
- In the explorer / resource list / overlay surfaces, `y` keeps its current contextual meaning (e.g. copy resource name in the pod list, copy yaml). No operator-pending there — there's no cursor-on-text, so vim semantics don't apply.
- `Y` is left alone where it has context-specific meaning (e.g. format-copy in ConfigMap/Label editors).
- Documented in `CHANGELOG.md` and the in-app `?`/F1 help.

## Alternatives considered

[...]

PR #185, opened 2026-05-08
feat(viewers): vim text-object selection (viw/vaw/viW/vaW)

sp71:
## Summary

Adds vim-style word/WORD text-object selection (`viw`, `vaw`, `viW`, `vaW`) to all five vim-style viewers (Log, YAML, Diff, Describe, Events). After entering visual mode, `i`/`a` arms a pending operator on the `Model`; the next `w`/`W` resolves into a character-wise selection covering the word under the cursor.

Also tightens the visual-yank status message: single-line char/block selections now say `Copied` instead of the misleading `Copied 1 line`. Multi-line and line-mode yanks keep their line counts.

## Type of change

- [x] feat: new user-facing feature

## Related issue

https://github.com/janosmiko/lfk/issues/154

## Changes

- New `Model.pendingTextObject byte` field tracks the pending operator (0 / `'i'` / `'a'`).
- New `consumeTextObjectPrelude` helper at the top of every visual-mode handler resolves `i`/`a` + `w`/`W` into a character-wise range (also clears stale `pendingG`).
- New range helpers in `update_vim.go`: `innerWordRange`, `aroundWordRange`, `innerWORDRange`, `aroundWORDRange`, plus a `textObjectRange` dispatcher. Boundary classification follows the existing `isWordBoundary`; WORD treats only space/tab as boundaries.
- Per-viewer `apply*TextObject` functions (Log, YAML, Diff, Describe, Events) commit the resolved range as a char-wise selection. YAML clamps the result past `yamlFoldPrefixLen` so the selection never enters the fold-prefix gutter.
- `formatVisualYank(clipText, mode, lineCount)` in `yank.go` returns `Copied` for single-line char/block yanks and `Copied N line(s)` otherwise. All five viewers' yank handlers use it.
- In-app help (`internal/ui/help_sections.go`) and `docs/keybindings.md` updated for all five viewers.

## Testing

- [x] `go build ./...` succeeds
- [x] `go test ./...` passes
- [x] `golangci-lint run` passes (via pre-commit hook)
- [x] Manually verified against a real cluster (when behavior changed)

## Checklist

- [x] Commits follow conventional commit style
- [x] README / `docs/` updated when user-visible behavior or config changed
- [x] `docs/keybindings.md` and the in-app help screen updated when keybindings changed
- [x] No secrets, tokens, or personal data included in diffs, logs, or screenshots
- [x] PR is opened from a feature branch on my fork (not from my fork's `main`)


sp71:
**The 4 `case "esc": m.pendingTextObject = 0` nits are not actionable** — they appear to miss the prelude. Every visual-mode handler calls `m.consumeTextObjectPrelude(key)` before the `switch` runs (see `internal/app/update_vim.go:263-274`):

```go
func (m *Model) consumeTextObjectPrelude(key string) (byte, string, bool) {
    if m.pendingTextObject == 0 {
        return 0, "", false
    }
    op := m.pendingTextObject
    m.pendingTextObject = 0   // ← unconditionally cleared when set
    m.pendingG = false
    if key == "w" || key == "W" {
        return op, key, true
    }
    return 0, "", false
}
```

So when `esc` reaches the switch, `pendingTextObject` is already `0`. Adding the assignment in each `case "esc":` arm would be dead code — that's exactly why an earlier round of this PR removed it from those four arms (squashed since). Coverage is exercised by `TestLogVisualEscClearsPendingOperator` which verifies the post-esc state is clean.

**Docstring coverage warning addressed in `b380200`** — the two unexported helpers (`innerRangeWith`, `aroundRangeWith`) now have short docstrings, which should push coverage above 80%.

PR #219, opened 2026-05-12
fix(scheduler): recover panics in task Fn so registry entries don't leak

janosmiko:
## Summary

User report: after disabling auto-watch and navigating to Pods, everything loads but the upper-right title-bar spinner keeps spinning. The user reports that opening the background-tasks overlay makes it stop.

## Root cause

\`runTask\` (\`internal/app/scheduler/scheduler_workers.go:267\`) calls \`task.req.Fn(ctx)\` inline without a panic recover. A panic in Fn unwinds the worker goroutine, skipping all four cleanup steps that follow:

1. \`r.unregisterRunning(...)\` — leaks an entry in \`r.runningTasks\` for that kube-context.
2. \`r.Finish(visID)\` — never called, so the registry entry stays with \`FinishedAt\` zero.
3. \`task.future <- Result{...}\` — never delivered, so callers blocking on the future leak.
4. \`close(task.future)\` — never closed.

\`lenLocked\` counts \"running\" entries (\`FinishedAt\` zero) without an expiry, so \`LenIndicator()\` never drops to zero and the title-bar spinner spins forever. The future leak strands the \`scheduleK8sCall\` closure on a channel receive forever, which also leaks a Bubble Tea goroutine.

A third side-effect: the worker goroutine itself dies, so the worker pool shrinks by one slot per panic. On the 4-worker prod default it's invisible until enough panics accumulate.

## Fix

Wrap the Fn call in \`safeRunFn\`, a tiny helper that \`defer recover()\`s the panic and converts it into a descriptive error (\`task panic: <value>\`). \`runTask\`'s existing branches then run their normal cleanup with err set to the panic error — the worker survives, the registry stays consistent, and the future resolves.

## Test plan

- [x] New test \`TestRunTask_RecoversPanicInFn\` submits a deliberately panicking Fn and asserts:
   1. The future resolves with a non-nil error containing the panic value.
   2. \`LenIndicator\` drops to zero after the linger window.
   3. A follow-up \`Submit\` still runs on the surviving worker pool.
   Pre-fix: test crashes the test process (Go propagates the unrecovered goroutine panic). Post-fix: test passes.
- [x] \`go test ./... -race -count=1 -short\` clean.
- [ ] Manual: user to confirm the spinner stops on navigation now that panicking tasks (likely \`GetAllPodMetrics\` or an event-list edge case on their cluster) clean up properly.

## Notes

The exact panic source is opaque to lfk — the symptom is the leak, not the K8s call itself. If the user can reproduce reliably, adding a one-line logger.Warn inside \`safeRunFn\` (something like \`logger.Warn(\"scheduler task panicked\", \"kind\", kind, \"name\", name, \"panic\", p)\`) would surface the offending call site. Left out of this PR to keep the diff minimal — happy to follow up if useful.

### DIFF
--- a/internal/app/scheduler/runtask_panic_test.go
+++ b/internal/app/scheduler/runtask_panic_test.go
@@ -1,0 +1,83 @@
+package scheduler
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestRunTask_RecoversPanicInFn locks in the recover-and-Finish semantics
+// the worker loop must apply when a task's Fn panics. Without the
+// recovery, the worker goroutine dies mid-flight and the registry entry
+// stays in r.tasks with FinishedAt zero — lenLocked counts it as a
+// running task forever, so LenIndicator never drops to zero and the
+// title-bar spinner spins endlessly. The future also stays unresolved,
+// so the caller's scheduleK8sCall closure blocks forever on the channel
+// receive. Both symptoms are user-visible (frozen spinner, leaked
+// goroutine).
+//
+// The fix wraps Fn in a defer-recover that converts a panic to an error,
+// then runs the normal cleanup (unregisterRunning, Finish, deliver to
+// future, close future) so the worker survives and the registry stays
+// consistent.
+func TestRunTask_RecoversPanicInFn(t *testing.T) {
+	r := New(0)
+	r.SetWorkersForTest(2, 0)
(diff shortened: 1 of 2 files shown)


janosmiko:
Closing: misdiagnosed the user's bug. The real fix is in #220 (title-bar spinner spinning during the 10s linger window). The panic-recovery code here is a real defense-in-depth improvement against any future Fn that panics, but lfk has no observed panics in production today and YAGNI applies. If a panic is ever reported the diff and test in this PR can be cherry-picked.

PR #223, opened 2026-05-12
feat(karpenter): first-class actions for NodePool / NodeClaim / EC2NodeClass

janosmiko:
## Summary

First batch of the planned Karpenter / Knative / OpenCost work. This PR ships Karpenter; Knative Serving follows next.

### What changes

- **NodeClaim** gains three Karpenter-native actions on top of the standard Describe/Edit/Delete/Events:
  - **Disrupt** — routed through the existing type-to-confirm overlay (user types \`DELETE\` + Enter) before the NodeClaim is deleted. Karpenter observes the deletion and tears down the underlying cloud instance plus the matching core Node.
  - **Cordon Node** / **Drain Node** — resolve \`status.nodeName\` from the NodeClaim and shell out \`kubectl cordon / drain\` against the resolved node. Drain carries the same \`--ignore-daemonsets --delete-emptydir-data\` flags as the standalone Node Drain action. An unbound NodeClaim (Karpenter still provisioning) returns a clear \"status.nodeName empty\" error instead of a generic failure.
- **NodePool** and **EC2NodeClass** stay on the standard Describe/Edit/Delete/Events surface. Disable/Enable via \`spec.disruption.budgets\` is deferred — patching budgets overlaps with user-managed config and deserves its own UX pass.
- **Icons** distinguish the node-family in the sidebar: NodePool ⏣, NodeClaim ⌬, EC2NodeClass ⌶. Core Node keeps ⌹.
- **Read-only mode** gates the three new mutating labels.

### Mechanics

- New \`internal/k8s/karpenter.go\` exposes \`GetNodeClaimNodeName\` and \`DisruptNodeClaim\` on the dynamic client.
- New \`internal/app/commands_karpenter.go\` holds the Tea commands plus the sub-dispatcher \`executeActionKarpenter\` that routes the new labels.
- \`update_actions_extended.go\` (new) extracts the existing \`executeActionExtended\` so \`update_actions.go\` stays under the 800-line file cap (it was at the boundary before this change).

### Out of scope (follow-ups when justified)

- Disable/Enable NodePool via \`spec.disruption.budgets\`.
- Azure / GCP NodeClass curation (fall through to generic CRD path).
- \"Go to Node\" navigation from a NodeClaim row.

## Test plan

- [x] \`internal/model/actions_karpenter_test.go\` — action-menu shape for all three kinds.
- [x] \`internal/k8s/karpenter_test.go\` — dynamic-client tests for \`GetNodeClaimNodeName\` (bound / unbound / not-found) and \`DisruptNodeClaim\` (success / not-found).
- [x] All existing tests still pass (\`go test -race ./...\` clean).
- [x] Manual smoke on a real Karpenter cluster (AWS recommended).

### DIFF
--- a/internal/app/commands_karpenter.go
+++ b/internal/app/commands_karpenter.go
@@ -13,8 +13,8 @@ import (
 )
 
 // executeActionKarpenter dispatches the Karpenter-specific action
-// labels (Disrupt, Cordon Node, Drain Node) into the matching Tea
-// command. Returns ok=true when the label is handled here so
+// labels (Disrupt, Cordon / Uncordon / Drain Node) into the matching
+// Tea command. Returns ok=true when the label is handled here so
 // executeActionExtended can early-return without growing its switch.
 // Disrupt opens the type-to-confirm overlay; the deferred mutation
 // lives in update_overlays_confirm.go's "Disrupt" branch.
@@ -25,6 +25,9 @@ func (m Model) executeActionKarpenter(actionLabel string) (tea.Model, tea.Cmd, b
 		return mdl, cmd, true
 	case "Cordon Node":
 		mdl, cmd := m.executeActionSimpleLoading("Cordoning node from NodeClaim", m.cordonNodeFromClaim)
+		return mdl, cmd, true
+	case "Uncordon Node":
+		mdl, cmd := m.executeActionSimpleLoading("Uncordoning node from NodeClaim", m.uncordonNodeFromClaim)
 		return mdl, cmd, true
 	case "Drain Node":
 		mdl, cmd := m.executeActionSimpleLoading("Draining node from NodeClaim", m.drainNodeFromClaim)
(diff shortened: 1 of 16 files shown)


janosmiko:
Addressed CodeRabbit feedback in e8d6004:

- **Empty error string fix** (`commands_karpenter.go`): when `kubectl` exits non-zero with empty stderr (cancelled context, kubeconfig missing on disk, etc.), the error message used to be `"cordon <node> (from NodeClaim X): "` with a trailing blank. Now falls back to the exec error's own message so users always see concrete failure details.

- **Negative test asserts** (`actions_karpenter_test.go`): `NodePool` and `EC2NodeClass` now explicitly `NotContains` `"Cordon Node"` / `"Drain Node"`, mirroring the existing `Disrupt` exclusion. Prevents a future refactor from silently re-introducing the node-bound verbs on the wrong kind.

- **Skipped: "remove emojis from Icon.Emoji"**: the `Icon` struct has dedicated `Unicode` / `Simple` / `Emoji` / `NerdFont` fields. The `Emoji` field is literally the emoji-rendering branch — every existing entry in `metadata.go` carries an emoji glyph there (`🖥️`, `🌍`, `🔒`, etc.). The project no-emoji rule applies to code / comments / docs, not to a data field whose sole purpose is to hold an emoji.

PR #224, opened 2026-05-12
feat(knative): first-class Knative Serving (Activate) + Eventing icons

janosmiko:
## Summary

Second of the three planned ecosystem PRs. Covers Knative **Serving** and **Eventing** in one batch:

- **Serving** gains one Knative-native verb on Revision plus per-kind icons across Service / Revision / Configuration / Route.
- **Eventing** gets curated metadata + kind-specific icons across Trigger / Broker / EventType / Channel / Subscription. The default action menu already covers Eventing's surface, so this is icons + metadata only — there's no Eventing-equivalent of \"send all traffic here.\"

### Serving — Activate

**Revision** action menu adds:

- **Activate** — patches the parent Knative Service's \`spec.traffic\` to route 100% of traffic to the selected Revision. Canonical Knative rollback / promotion gesture. Parent Service resolved via the standard \`serving.knative.dev/service\` label that Knative sets on every Revision; an orphaned Revision (no such label, rare — usually mid-deletion) returns a clear error naming the missing label instead of attempting a blind patch.

The patch replaces \`spec.traffic\` with a single entry pinned to the chosen Revision (\`latestRevision: false\`). JSON merge-patch treats arrays as atomic, which is the intended semantic — \"this Revision gets all the traffic, drop everything else.\"

**Configuration** and **Route** stay on the standard surface; their CRD printer columns (URL, LatestReady, Ready) already carry the operationally useful state.

**Knative Service** (Kind=\"Service\") collides with the core \`/v1/Service\` handler. Because the core Service menu (Tail Logs, Port Forward, etc.) is largely applicable to Knative Services too — they expose pods underneath — Knative Services deliberately stay on the core menu. A Knative-specific Service overlay (traffic-split editor, scale-min/max annotations) is deferred to a follow-up that distinguishes Service-by-APIGroup.

### Eventing — icons only

Curated entries:

| Resource | Glyph | Note |
|---|---|---|
| \`eventing.knative.dev/triggers\` | ↯ | Lightning Down — narrow variant; ⚡ renders at width 2 and breaks single-cell sidebar |
| \`eventing.knative.dev/brokers\` | ❖ | Central event hub |
| \`eventing.knative.dev/eventtypes\` | ❑ | Typed-event schema registry (new curated entry) |
| \`messaging.knative.dev/channels\` | ⇉ | Event pipe (new curated entry) |
| \`messaging.knative.dev/subscriptions\` | ◎ | Channel listener (new curated entry) |

Triggers / Brokers were already in \`BuiltInMetadata\` with the generic ⧫ CR glyph; this swaps them to kind-specific glyphs that match the sidebar pattern. EventTypes / Channels / Subscriptions are new curated entries.

### Icons (Serving)

Per-kind glyphs distinguish the four-piece Service / Revision / Configuration / Route mental model: ⌘ ⟲ ⌗ ↦. (⚙ and ⊞ were candidates but collide with admission configs / PVCs respectively, so swapped before commit.)

### Read-only mode

Gates Activate. Eventing additions are icons-only, no new mutating verbs.

### Mechanics

- New \`internal/k8s/knative.go\` exposes \`ActivateKnativeRevision\` on the dynamic client (with a 30 s \`context.WithTimeout\` since \`trackBgTask\` doesn't propagate the scheduler's per-task timeout).
- New \`internal/app/commands_knative.go\` holds the Tea command plus the \`executeActionKnative\` sub-dispatcher.
- \`update_actions_extended.go\` extracted from \`update_actions.go\` so the file stays under the 800-line cap. Same extraction trick PR #223 (Karpenter) uses; whichever PR merges first sets up the shared file and the other rebases cleanly.

### Out of scope (follow-ups when justified)

- Traffic-split overlay on Service (multi-Revision edit).
- Scale Min/Max via \`autoscaling.knative.dev/min-scale\` and \`max-scale\` annotations.
- Service-by-APIGroup dispatch so Knative-specific Service verbs can coexist with the core Service menu.
- InMemoryChannel and other messaging implementations.
- Source CRDs (ApiServerSource, PingSource, ContainerSource, KafkaSource, etc.) — many shapes; benefit from per-Source curation.
- Flow CRDs (Sequence, Parallel).
- Eventing-specific verbs (resend-from-DLQ, test-filter, show-broker-URL) — none are obvious wins and all require significant new UI surface.

## Test plan

- [x] \`internal/model/actions_knative_test.go\`:
   - Serving: \`TestKnativeActions_Revision\` / \`_Service\` / \`_Configuration\` / \`_Route\` — \`Activate\` present on Revision, absent everywhere else.
   - Eventing: \`TestKnativeEventingActions\` — Broker / Trigger / Channel / Subscription / EventType each surface the standard Describe/Edit/Delete/Events set, and \`Activate\` is explicitly absent.
- [x] \`internal/k8s/knative_test.go\` — dynamic-client tests for \`ActivateKnativeRevision\` (happy path collapsing traffic to one entry, orphaned Revision, Revision-not-found).
- [x] All existing tests still pass (\`go test -race ./...\` clean).
- [ ] Manual smoke on a real Knative cluster.

### DIFF
--- a/internal/app/commands_knative.go
+++ b/internal/app/commands_knative.go
@@ -1,0 +1,44 @@
+package app
+
+import (
+	"fmt"
+
+	tea "github.com/charmbracelet/bubbletea"
+
+	"github.com/janosmiko/lfk/internal/app/scheduler"
+	"github.com/janosmiko/lfk/internal/logger"
+)
+
+// executeActionKnative dispatches Knative Serving action labels to
+// their Tea commands. Returns ok=true when the label is handled here
+// so executeActionExtended can early-return without growing its
+// switch. Currently surfaces only the Activate verb on Revision; the
+// traffic-split overlay on Service is deferred to a follow-up.
+func (m Model) executeActionKnative(actionLabel string) (tea.Model, tea.Cmd, bool) {
+	if actionLabel == "Activate" {
+		mdl, cmd := m.executeActionSimpleLoading("Activating Revision", m.activateKnativeRevision)
+		return mdl, cmd, true
+	}
+	return m, nil, false
+}
+
+// activateKnativeRevision patches the Revision's parent Knative
+// Service so 100% of traffic routes to the selected Revision. Wrapped
+// in trackBgTask so the title-bar indicator surfaces the in-flight
+// patch and the :tasks overlay records it. The parent Service name is
+// resolved by Client.ActivateKnativeRevision via the standard
+// serving.knative.dev/service label on the Revision; an orphaned
+// Revision returns a clear error naming the missing label.
+func (m Model) activateKnativeRevision() tea.Cmd {
+	ctx := m.actionCtx.context
+	ns := m.actionCtx.namespace
+	name := m.actionCtx.name
+	logger.Info("Knative Revision activation requested", "context", ctx, "namespace", ns, "name", name)
+	return m.trackBgTask(scheduler.KindMutation, "Activate Revision: "+name, bgtaskTarget(ctx, ns), func() tea.Msg {
+		parent, err := m.client.ActivateKnativeRevision(ctx, ns, name)
+		if err != nil {
+			return actionResultMsg{err: err}
+		}
+		return actionResultMsg{message: fmt.Sprintf("Activated Revision %s on Service %s (100%% traffic)", name, parent)}
+	})
+}
--- a/internal/app/readonly.go
+++ b/internal/app/readonly.go
@@ -79,6 +79,10 @@ var mutatingActions = map[string]bool{
 	"Uncordon Node": true,
 	"Drain Node":    true,
 
+	// Knative Serving mutations. Activate patches the parent Service's
+	// spec.traffic to send 100% to a Revision.
+	"Activate": true,
+
 	// Helm release mutations.
 	"Edit Values": true,
 	"Upgrade":     true,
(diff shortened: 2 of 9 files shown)


janosmiko:
Addressed CodeRabbit feedback in b781cb1:

- **Bounded API calls** (`knative.go`): `ActivateKnativeRevision` now bounds its GET-Revision and PATCH-Service calls with a 30 s `context.WithTimeout` (matching `scheduler.DefaultRequestTimeout`). `trackBgTask` wraps the inner function as `func() tea.Msg` with no context parameter, so the scheduler's per-task timeout doesn't propagate into mutation client methods — without this guard a hung apiserver would block the worker indefinitely.

- **Skipped: "remove emojis from Icon.Emoji"**: same false positive as on #223. The `Icon` struct has dedicated `Unicode` / `Simple` / `Emoji` / `NerdFont` fields. The `Emoji` field is literally the emoji-rendering branch — every existing entry in `metadata.go` carries an emoji glyph there. The project no-emoji rule applies to code / comments / docs, not to a data field whose sole purpose is to hold an emoji.

janosmiko:
Addressed the new CodeRabbit finding in 4a5de26:

- **gofumpt on metadata.go:359** — already fixed in f3ee513 (blank line between Serving and Eventing icon blocks; my local pre-commit's gofumpt didn't catch it but CI's v2.12.1 did).
- **Bare glyphs in the Knative comments** — quoted them in line with the Karpenter PR's pattern (`Core Node uses "⌹"` style). The glyphs themselves still live in the `Icon` literals immediately below; only the explanatory comment text changes.

PR #256, opened 2026-05-19
feat: add jump-back navigation history (#249)

### LINKED ISSUE #249: [feat]: Need ability to navigate back to individual resource pane when using `Owner` shortcut
### Problem / motivation

It would be nice to be able to have navigate/backforth using arrow keys when jumping to resource `owner` (via `o` shortcut)

### Proposed solution

Add ability to use right arrow to go back to previuos screen after pressing `o` shortcut

### Alternatives considered

Going back to pod view by going to resource type view and finding correct resource type

### Area

Navigation / layout

### Additional context

_No response_

### Pre-flight checks

janosmiko:
## Summary

Resolves #249.

Non-hierarchical "teleport" jumps previously discarded the user's origin, leaving no way back to the resource pane they came from. This adds a per-tab jump-back history.

- **`backspace`** — return to where you teleported from (full nav state, cursor, and filter restored).
- Tracked teleports: jump-to-owner (`o`), open port-forwards, orphans-overlay jump, bookmark jumps.
- Each teleport pushes a navigation snapshot onto a per-tab stack (capped at 50).
- Hierarchical `left`/`h`/`right`/`l` navigation is unchanged.
- Snapshots are fully deep-copied; restoring a stale target (context no longer valid) falls back to the cluster picker with a status message instead of failing.

Forward navigation was intentionally not included — `backspace` alone solves the filed issue, and reliable cross-terminal forward keys proved problematic.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race ./internal/app/... ./internal/ui/...`
- [x] New tests in `jump_history_test.go`: snapshot capture/restore fidelity, teleport+back returns to origin, empty-stack no-op, stack cap at 50, multi-level history, stale-target fallback, deep-copy isolation, per-tab independence
- [ ] Manual: press `o` on a pod, then `backspace` — returns to the pod pane

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,7 @@ Namespaces are **not** a navigation level. The current namespace is shown in the
 | `0` / `1` / `2` | Jump to clusters / types / resources level |
 | `J` / `K` | Scroll preview pane down/up |
 | `o` | Jump to owner/controller of selected resource |
+| `Backspace` | Jump back through teleport history (owner / port-forward / orphan / mark jumps) |
 
 ### Views and Modes
 
--- a/docs/keybindings.md
+++ b/docs/keybindings.md
@@ -21,6 +21,7 @@ Complete list of all keybindings in `lfk`. All keybindings can be overridden in 
 | `0` / `1` / `2` | Jump to clusters / types / resources level |
 | `J` / `K` | Scroll preview pane down/up |
 | `o` / `O` | `o` jumps to the owner/controller of the selected resource; `O` opens the cluster-wide orphan overview overlay |
+| `Backspace` | Jump back through teleport history (owner, port-forward, orphan, and mark jumps push history; hierarchical `h`/`l` navigation does not) |
 
 ## Views and Tools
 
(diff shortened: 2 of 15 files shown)


janosmiko:
Addressed all CodeRabbit findings in b4a65ec:

1. **Stale-context fallback cancellation** — Fixed. Added `cancelAndReset()` + `requestGen++` to the fallback branch. (The normal restore path already did this.)
2. **Bookmark jump history** — Fixed. Moved `pushJumpHistory()` from `jumpToSlot` into `navigateToBookmark`, after successful resolution — so the overlay Enter path records history too, and failed jumps record none.
3. **Backspace during text input** — Verified the reported path does not actually reproduce: `filterActive`/`searchActive` are routed to their input handlers in `update_keys.go` *before* `handleExplorerKey` is reached, so jump-back never sees those keystrokes. Added a defensive guard in `handleExplorerJumpKey` anyway.
4. **Test cap duplication** — Fixed. `TestJumpHistoryStackCap` now exercises the real `pushJumpHistory` path.

Added tests covering #2 and #3.

janosmiko:
@coderabbitai review

PR #291, opened 2026-05-28
fix: clear hint-bar message immediately on explorer navigation

janosmiko:
## Summary

Status and error toasts in the explorer hint bar lingered for the full 5s `scheduleStatusClear` timer. The startup tip already cleared on any keypress, but regular messages did not — so a `Read-only mode: ON` or error toast stayed on screen while the user navigated away.

Now any explorer **navigation** key clears the message immediately — anything that changes what the user is looking at:

- cursor movement: `h/j/k/l` + arrow keys
- top/bottom/home/end jumps: `g`/`G`/`Home`/`End`
- page scrolling: `ctrl+d`/`ctrl+u` (half), `ctrl+f`/`ctrl+b`/`PgDn`/`PgUp` (full)
- preview-pane scroll: `J`/`K`
- level jumps: `0`/`1`/`2`
- owner/back jumps: `o` / `Backspace`
- search-match jumps: `n`/`N`
- tab switching: `[`/`]`/`t`

`clearStatusOnNavigationKey` (in `internal/app/status_clear.go`) is called from `handleExplorerKey` *before* dispatch, so a handler that sets a fresh message for the same key (e.g. `JumpOwner` reporting "no owner", or `NewTab` rejected in union mode) still overwrites the cleared state and wins. The detection lives in a helper rather than inline in `handleExplorerNavKey` to keep that function under the gocyclo-30 cap.

Action keys that aren't navigation (e.g. `x`, `f`, `Enter`) leave the message intact.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race ./...`
- [x] `make lint`
- [x] `TestHandleKeyNavigationClearsStatusMessage` — integration via handleKey (j/k/h/l, arrows, g, G, End)
- [x] `TestClearStatusOnNavigationKey` — unit-tests the helper across the full nav/scroll set, plus negative cases (x/f/Enter keep the message)

### DIFF
--- a/internal/app/status_clear.go
+++ b/internal/app/status_clear.go
@@ -1,0 +1,35 @@
+package app
+
+import (
+	"time"
+
+	tea "github.com/charmbracelet/bubbletea"
+
+	"github.com/janosmiko/lfk/internal/ui"
+)
+
+// clearTransientStatus immediately drops any active status/error message so it
+// does not linger in the hint bar while the user navigates. The startup tip is
+// dismissed separately in handleKey. No-op when nothing is showing.
+func (m *Model) clearTransientStatus() {
+	if m.statusMessage == "" {
+		return
+	}
+	m.statusMessage = ""
+	m.statusMessageErr = false
+	m.statusMessageExp = time.Time{}
+}
+
+// clearStatusOnNavigationKey clears a lingering hint-bar message when msg is an
+// explorer movement key (up/down/left/right plus the top/bottom jumps), so a
+// toast disappears on the next navigation keystroke instead of waiting out its
+// 5s timer. Returns m unchanged for any other key.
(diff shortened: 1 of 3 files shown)


janosmiko:
Addressed the CodeRabbit nitpick: the integration table now includes `g` and `End`, and a new `TestClearStatusOnNavigationKey` unit test exercises the full navigation/scroll key set directly. Also expanded the feature itself per reviewer (human) request — page scroll (ctrl+d/u, ctrl+f/b, pgdn/pgup), preview scroll (J/K), level jumps (0/1/2), owner/back jumps (o/backspace), search-match jumps (n/N), and tab switching ([/]/t) now all clear a lingering hint-bar message.

PR #294, opened 2026-05-29
fix: cluster dashboard rendering, scrolling, and responsive usage bars (#293)

### LINKED ISSUE #293: [bug]: The cluster view field backgrounds don't honour the theme
### Describe the bug

In WezTerm and iTerm2 if you pick a theme with a background colour that isn't black the Cluster page 'tears' bits of its fields with a black background instead of the theme background colour.

### Steps to reproduce

1. Run in iTerm2 or WezTerm (may be others too)
2. shift+T pick a theme with a non-black background
3. Navigate to a cluster's details page
4. Enter full screen mode (right or shift+F)

### Expected behavior

The background should stay solid

### Actual behavior

Bits of the UI suddenly have a black background. It doesn't exhibit this behaviour when the cluster details are in the right most pane.

### lfk version

0.12.7 (commit: d825f6b, built: 2026-05-28T04:42:12Z)

### Installation source

Homebrew

### Operating system

macOS 26.5 (25F71)

### Terminal emulator

iTerm2 / WezTerm

### kubectl version

Client Version: v1.36.1

### Kubernetes server version

v1.35.2

### Relevant config

```yaml

```

### Relevant log output

```shell

```

### Additional context

Image

Image

### Pre-flight checks

janosmiko:
## Summary

Fixes #293 and a cluster of cluster-dashboard issues found while testing it. Four commits, all scoped to the fullscreen cluster dashboard:

1. **Theme background fill (#293).** The two-column dashboard never ran content through `FillLinesBg`, so under non-black themes styled spans/padding rendered with the terminal default (black) — "tearing" the panel. Now mirrors the single-column path.

2. **Row width fit.** Fixing (1) exposed a latent miscalc: rows were built to `fullW` but `ActiveColumnStyle`'s padding leaves only `fullW-2`, so once padding became styled lipgloss wrapped every row. Columns now allocate within `fullW-2`.

3. **Scroll clamp.** `clampPreviewScroll` bounded against the right-column preview (narrow width + `previewEventsContent`), letting the dashboard scroll past its last line. Added `clampDashboardScroll`, routed to whenever `fullscreenDashboard` is active.

4. **Responsive usage bars.** Bars were a fixed 30 cols (15/node) baked at load time, so they were too narrow in fullscreen. The dashboard data is now retained per context and composed at render time with width-aware bar/separator widths — wide in fullscreen, compact in the right pane, clamped so nothing overflows the two-col left pane. Recomposed on load, fullscreen toggle, and resize.

Dashboard rendering + scroll helpers were extracted to `view_dashboard.go` to keep `view.go` under the 800-line cap.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go test -race ./...`, `make lint`
- [x] `TestViewExplorerDashboardTwoColFillsThemeBackground` — bg fill + no row-overflow wrap
- [x] `TestClampPreviewScrollDashboard` — over-scroll bounded
- [x] `TestDashboardBarsScaleWithWidth` — fullscreen bars wider than right-pane and the old fixed 30
- [x] `TestComposeDashboardFitsContentWidth` — no composed line exceeds the content width (wrap guard) across single-col / two-col / right-pane


janosmiko:
@coderabbitai The node-name wrap concern is already handled. CodeRabbit reviewed commit 6ac3622; the follow-up commit 9bf21e2 wraps both per-node rows in `ansi.Truncate(..., w.content, "")`, so the `statusDot + name + roleStr` line (and the CPU/MEM line) are clipped to the actual column width and can no longer wrap — regardless of the 48-char raw cap.

Hardened the regression test (commit f7d5667): `TestComposeDashboardFitsContentWidth` now feeds a 51-char node name and asserts every composed line stays within the content width in fullscreen single-col, two-col, and the narrow right pane. Verified it fails without the name-line truncation and passes with it.

PR #310, opened 2026-05-31
feat: per-kind sort memory and per-context column memory

### LINKED ISSUE #302: [feat]: Remember sort preference per resource kind
### Problem / motivation

Whenever you get in to the resource list the sorting is always by the metadata.name, and it's cool you can change the sort order but when you leave that resource and come back to it later (or immediately) the sorting has reset.

### Proposed solution

Save the sort order - if not on disk then at least in memory while lfk is running.

### Alternatives considered

_No response_

### Area

Resource list / preview

### Additional context

_No response_

### Pre-flight checks

janosmiko:
## Summary

Two related session-memory improvements for the resource list, so user choices stick across navigation instead of resetting.

- **Remember sort preference per resource kind (#302).** Sorting a list then leaving and returning reset the sort to the default. Now the chosen sort column + direction are remembered per resource kind, scoped by cluster context, for the running session. The `-` reset clears the remembered choice. Priority: remembered session sort → configured view default → built-in `Name` ascending. Stored in an in-memory map mirroring the existing `cursorMemory`, deep-copied per tab.
- **Scope column visibility memory per cluster context.** Column visibility / hidden built-ins / column order were already remembered per kind, but shared across all clusters. They are now keyed by `context + kind` (mirroring the sort memory), so the same kind can show different columns in different clusters. Reset clears only the current context's config.

Both features are in-memory (session-only) and mirror each other.

## Implementation notes

- New `columnMemoryKey(kind)` routes every read (`view.go`) and read/write (`update_column_toggle.go`) of `sessionColumns` / `hiddenBuiltinColumns` / `columnOrder` through one context-scoped key.
- Sort recording added at all 6 user sort-change sites (`>` `<` `=` keys, header click, `:sort`); reset uses a dedicated `forgetSort`.
- Extracted `tabs_copy.go` and `update_keys_sort.go` to keep files under the 800-line lint limit.

## Test plan

- [x] New unit tests: per-kind/per-context isolation, memory-wins-over-view, reset-clears, round-trip (`app_sort_memory_test.go`, `update_column_toggle_context_test.go`).
- [x] Existing column-toggle tests updated to the new keying.
- [x] `go build ./...`, `go vet`, `gofmt -l` clean.
- [x] `go test ./...` passes (incl. `-race`).
- [x] `golangci-lint run ./...` → 0 issues.

Closes #302


janosmiko:
@coderabbitai review

janosmiko:
Addressed both review comments in 3340464 and 20b2975:
- `app_sort_memory_test.go`: `ui.BuildView` errors now fail the test via `t.Fatalf` (both fixtures).
- `copyItemCache`: now clones each Item's `Columns`/`Conditions`/`GroupedRefs` (pure value-struct slices) so a tab's cached snapshot can't be mutated through a shared backing array by the in-place appends in the metrics / event-grouping paths. `Raw` is intentionally left shared and documented — it's the immutable source object, only re-read for JSONPath.

PR #318, opened 2026-05-31
fix: attribute cluster credential failures + Prometheus pod metrics + log viewer cursor

janosmiko:
Three related fixes surfaced while triaging confusing log output on a multi-cluster setup.

## 1. Attribute credential/SSO failures to their cluster (`fix(logs)`)
**Problem:** `20:55 [ERR] stderr: aws: [ERROR]: The SSO session ... has expired` — no way to tell *which* cluster's SSO expired.

client-go runs kubeconfig exec credential plugins with stderr hardwired to `os.Stderr` and caches authenticators globally, so the raw line carries no context and the text never names the AWS profile.

- Wrap each context's REST transport (built via `rest.HTTPClientFor` + `NewForConfigAndClient`, so the wrapper is outermost — around client-go's exec auth). On a credential failure it emits one deduped, actionable line: cluster context + AWS profile + `aws sso login --profile X` hint, to both the file log and the in-app overlay.
- Demote the now-redundant contextless exec-plugin stderr line to Debug and keep it out of the overlay. Safe: kubectl subprocesses capture their own stderr, so the global capture only sees in-process exec plugins.

## 2. Prometheus fallback for pod metrics (`feat(metrics)`)
**Problem:** `pod metrics list load failed ... metrics API unavailable` on a cluster where pod metrics *are* available via Prometheus.

Pod metrics only ever queried `metrics.k8s.io` (node metrics already had a Prometheus path). Added a per-pod Prometheus query and routed `GetPodMetrics`/`GetPodsMetrics`/`GetAllPodMetrics` through the same Prometheus↔metrics-API fallback. The warning now fires only when **every** source fails and reads `pod metrics unavailable: no metrics-server or Prometheus source`.

Also keyed `promSvcCache` by context name instead of the clientset value — `clientsetForContext` builds a fresh clientset per call, so the old key never hit and re-ran Prometheus service discovery on every poll.

## 3. Preserve log viewer cursor on debug toggle (`fix(logviewer)`)
Pressing `d` reset scroll+cursor to the top. Now anchors on the selected entry's identity across the filter change; if the selected line is a DBG entry being hidden, falls back to the nearest surviving entry.

## Test plan
- [x] TDD: new unit tests for credential tagging, stderr demotion, cursor anchoring, Prometheus pod queries/routing
- [x] `go build ./...`, full `go test ./...`, `-race` on touched packages
- [x] `golangci-lint` clean
- [x] security-reviewer + go-reviewer on the auth diff; HIGH (nil-transport) and defensive findings applied
- [x] docs/config-reference.md updated (`node_metrics` now governs pod metrics too)

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -117,7 +117,7 @@ Each monitoring entry accepts the following top-level fields:
 |---|---|---|---|
 | `prometheus` | object | *(auto-discovery)* | Prometheus endpoint configuration. See endpoint fields below. |
 | `alertmanager` | object | *(auto-discovery)* | Alertmanager endpoint configuration. See endpoint fields below. |
-| `node_metrics` | string | *(auto-detect)* | Node metrics source: `"prometheus"` (use Prometheus queries), `"metrics-api"` (use metrics.k8s.io API). When empty, uses Prometheus if a prometheus endpoint is configured, otherwise falls back to metrics-api. |
+| `node_metrics` | string | *(auto-detect)* | Metrics source for node **and pod** usage: `"prometheus"` (use Prometheus queries), `"metrics-api"` (use metrics.k8s.io API). When empty, uses Prometheus if a prometheus endpoint is configured, otherwise metrics-api. Either way the other source is tried as a fallback, so pod metrics resolve on clusters served only by Prometheus. |
 
 ### Monitoring Endpoint Fields
 
--- a/internal/app/commands_load_metrics.go
+++ b/internal/app/commands_load_metrics.go
@@ -184,7 +184,7 @@ func (m Model) loadPodMetricsForList() tea.Cmd {
 			metrics, err := client.GetAllPodMetrics(ctx, kctx, ns)
 			if err != nil {
 				logger.WarnOnce("pod-metrics-list-load", kctx+"/"+ns,
-					"pod metrics list load failed", "context", kctx, "namespace", ns, "error", logger.Redact(err.Error()))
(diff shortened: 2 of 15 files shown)


janosmiko:
Thanks @coderabbitai — addressed in 5b9b4b6.

**1. `isExecCredentialError` "executable not found" form (auth_tagging.go) — already handled.**
The `getting credentials: exec: executable <name> not found` error is caught by the first branch, `strings.Contains(s, "getting credentials")`, since client-go prefixes all exec credential failures (both "not found" and "failed with exit code N") with `getting credentials:`. Added an explicit unit-test case for the not-found form to lock this in.

**2. Three near-identical Prometheus vector parsers — consolidated.**
Extracted `parsePrometheusVector(data, keyFunc)`; the node/pod/container parsers are now thin key-extraction closures. Existing parser tests still pass.

**3. Empty Prometheus result not falling back to metrics-API — intentionally not changed.**
Treating an empty (HTTP 200, `status:success`, `result:[]`) response as a failure that triggers fallback would regress the legitimate case: on a Prometheus-only cluster, a namespace with no pods returns empty, which would then fall through to the unavailable metrics-API and raise a false `no pod metrics source available` warning for a namespace that simply has no pods. Empty-vs-absent can't be distinguished cheaply, and the node-metrics path behaves identically (consistency). Users whose real source is metrics-server can set `node_metrics: metrics-api`, which tries that first. Fallback still fires correctly when the Prometheus query itself errors (proxy/network failure).

PR #320, opened 2026-05-31
refactor(app): group YAML viewer fields into yamlViewState

sp71:
## Summary

First step of decomposing the massive `Model` object (`internal/app/app.go`: 423 fields, ~1248 methods). This PR extracts the full-screen YAML viewer's 17 flat `yaml*` fields into a single embedded `yamlViewState` value (`423 -> 407` fields), following the sub-state pattern already established in the codebase (`canIState`, `whoCanState`, `syncWaveState`, etc.).

It is intentionally scoped as a **pilot** to validate the approach before I apply it to the other large field clusters. If you're happy with the direction, I'm glad to do the rest:

- `diffViewState` (~23 fields, `update_diff.go`)
- `describeViewState` (~16 fields, `update_describe.go`)
- `logViewState` (~44 fields, `update_logs*.go`)

Each would be an independent, behaviour-preserving PR using the same recipe.

## Type of change

- [x] refactor: code change that neither fixes a bug nor adds a feature

## Changes

- New `yamlViewState` (`internal/app/yamlview.go`) grouping content, scroll/cursor, search, visual selection, wrap, and collapsible sections; with a `copy()` that deep-copies `matchLines`/`sections`/`collapsed`.
- Mechanical `m.yamlX` -> `m.yamlView.x` rename across `update_yaml*.go`, `view_yaml.go`, `view_right.go`, `update_navigation.go`, and tests.
- `tabs.go` save/restore/clone keep `TabState`'s flat `yaml*` fields; only the `Model` side is regrouped, so per-tab persistence is byte-for-byte unchanged.
- `yaml`-prefixed `Model` methods (`yamlViewportLines`, `yamlCursorCol`, etc.) are untouched; only field accesses moved.
- Added `TestTabSwitchPreservesYAMLViewerState` (pins the save/switch/restore contract) and `yamlViewState.copy()` unit tests.

## Testing

- [x] `go build ./...` succeeds
- [x] `go test ./...` passes
- [x] `golangci-lint run` passes
- [x] Manually verified against a real cluster (when behavior changed)

Manual verification on a local `kind` cluster (full-screen YAML on a Deployment): scroll (`j`/`k`/`Ctrl+D`/`G`), collapse/expand all folds (`Z`), visual line selection (`V`), wrap (`Ctrl+W`), and a tab round-trip (collapse -> new tab -> switch back) confirming the YAML view + collapsed state restore intact.

Also confirmed the production diff is a pure rename: inverting `m.yamlView.x` -> `m.yamlX` across the changed logic files produces a zero-line diff against `main`.

## Checklist

- [x] Commits follow conventional commit style (`feat:`, `fix:`, `refactor:`, `docs:`, `test:`, `chore:`, `perf:`, `ci:`)
- [x] README / `docs/` updated when user-visible behavior or config changed
- [x] `docs/keybindings.md` and the in-app help screen updated when keybindings changed
- [x] No secrets, tokens, or personal data included in diffs, logs, or screenshots
- [x] PR is opened from a feature branch on my fork (not from my fork's `main`)


sp71:
This PR is intentionally a pure, behaviour-preserving rename (inverting `m.yamlView.x` → `m.yamlX` over the changed logic files yields a zero-line diff against `main`), so I'm keeping pre-existing behaviour and pre-existing inefficiencies out of scope to preserve that property.

- **`app.go` stale `yamlContent` comment** — fixed in f487d18 (the rename removed the flat field, so the comment was genuinely stale). Comment-only.

- **`tabs.go` shared `matchLines` backing array** — pre-existing: `main` does the same `t.yamlMatchLines = m.yamlMatchLines`. In practice it's safe, because `matchLines` is only ever reassigned (`= nil`, then `= append(nil-slice, …)` on each search recompute), never mutated in place through the shared array — so a saved tab's slice is never written through. Leaving the persistence semantics byte-identical here. Note the new `yamlViewState.copy()` already deep-copies `matchLines`/`sections`/`collapsed`; a sensible follow-up is to route `saveCurrentTab`/`loadTab` through `copy()`, which closes this for all fields at once — happy to do that as a separate change.

- **`update_yaml.go` duplicate `buildVisibleLines` in `yamlScrollToMatchFolded`** — pre-existing on `main` (and a couple of other call sites it didn't flag). Genuine micro-opt, but it's not a rename; I'd rather not expand this PR's scope and break the zero-diff guarantee. Can fold it into a small follow-up cleanup PR.

The docstring-coverage pre-merge warning is repo-wide and not specific to this diff

janosmiko (approved):

PR #332, opened 2026-06-01
feat(security): finding-ignore visibility — fix dead toggle, namespace scope, config globs

janosmiko:
Implements the security-finding visibility TODO cluster (L955–L957). Investigation found L955/L956 already shipped in #183 (checkboxes just never ticked) — and that the show-ignored toggle was **broken**.

## Changes

**fix(security): make the show-ignored toggle reachable (`ctrl+i` was dead)**
- `ctrl+i` is byte-identical to `Tab` (0x09); bubbletea emits `"tab"`, never `"ctrl+i"`, so the binding could never fire. Handler unit tests passed only because they bypassed the key string.
- Rebound `SecurityIgnoreToggle` to `i`, dispatched via a new gated `handleExplorerSecurityViewKeys` so it shadows the (meaningless) Label Editor only on security views — same idiom as `ClusterColorPicker` reusing the Logs key at the cluster picker.
- Added the toggle to the in-app help (was missing) + `keybindings.md`.
- Guard test `TestDefaultKeybindingsNoUnreachableControlAliases` rejects the whole dead class (`ctrl+i`/`ctrl+m`/`ctrl+[`).

**feat(security): namespace-scoped ignores + config-file glob patterns (L957)**
- `Namespace` field on `SecurityIgnoreRule`; scope precedence cluster-wide group > namespace > exact resource. New "Ignore (Namespace)" action; un-ignore peels the most specific rule first and warns when a broader rule still hides the row.
- New top-level `security.ignore_patterns` (cluster/source/group/namespace globs + comment) applied at load. The filter consults interactive YAML rules + config patterns together; the action menu reads YAML-only so config-only ignores never offer a misleading "Un-ignore". The `i` toggle still reveals config-hidden findings.
- Slash-agnostic `globMatch` (`*`/`?`); all-empty patterns dropped; per-cluster `ignore_patterns` warns (use the per-pattern `cluster` glob).

A multi-agent adversarial review raised 18 findings (16 confirmed); all real ones are addressed in this branch, including the 2 HIGH test-coverage gaps. One LOW enhancement (per-row ignored markers in show-ignored mode) is deferred.

## Test plan
- [x] `go build ./...`, `go vet ./...` clean
- [x] `go test -race ./...` green (+31 new tests: glob matching, namespace-scope predicates, checker combination, cluster-scoped findings, action menu + dispatch gating, config load, and `groupFindings` driven end-to-end through a real checker incl. show-ignored tagging)
- [ ] Manual: open Security → a source → a finding group → drill in; `x` shows Ignore (Group/Namespace/This Resource); `i` toggles ignored visibility on the security view (and is a no-op elsewhere)
- [ ] Manual: add a `security.ignore_patterns` entry; confirm matching findings hide and `i` reveals them

### DIFF
--- a/docs/config-example.yaml
+++ b/docs/config-example.yaml
@@ -87,6 +87,17 @@ icons: auto
 #   sources:
 #     falco: false           # disable a single source; omitted = enabled
 #     trivy: false
+#   # Declarative, glob-based ignore rules (applied at load). Each field is a
+#   # glob (* and ?); an empty field matches anything; a finding is hidden when
+#   # every non-empty field matches. Read-only — the `i` toggle still reveals
+#   # them. Interactive per-finding ignores (action menu `x`) are stored
+#   # separately in ~/.local/state/lfk/security_ignores.yaml.
+#   ignore_patterns:
+#     - source: trivy-operator   # source id; "" = any
+#       group: "CVE-2024-*"      # CVE id / check label / rule name; "" = any
+#       namespace: kube-system   # "" or "*" = any namespace (hides whole group)
+#       cluster: "prod-*"        # kube-context glob; "" = any cluster
+#       comment: accepted in system namespaces
 
 # ─── Log Path ──────────────────────────────────────────────────────────────────
 # Application log file location. Default: ~/.local/share/lfk/lfk.log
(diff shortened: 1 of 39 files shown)


janosmiko:
@coderabbitai review

PR #338, opened 2026-06-01
feat: hide individual resource types per cluster (#321)

### LINKED ISSUE #321: [feat]: Rarely used resource types and hiding
### Problem / motivation

H - Toggle rarely used resource types.

Brilliant, I love it. But there are resource types I rarely use that aren't in the list. I'd like to hide things like Ingresses - I use Gateways and xRoutes instead. I'd like to hide LimitRanges, Helm.Releases, etc, etc.

### Proposed solution

I don't think lowercase h is used yet - I'd propose that 'h' hides or unhides a resource type, while H shows you the hidden resource types. Keep the existing set of rarely used resource types because it's pretty good, but this way my local config can be adapted to my needs. I'd also like it if it recorded this configuration per-cluster because different clusters have different configuration needs.

### Alternatives considered

_No response_

### Area

Navigation / layout, Resource list / preview

### Additional context

_No response_

### Pre-flight checks

janosmiko:
## Summary

Closes #321. Adds the ability to **hide individual resource types** from the sidebar, persisted per cluster. Modeled on the existing pin subsystem; reveal folds into the existing `Shift+H` toggle.

- **Hide/show** a resource type via the **action menu (`x`)** at the resource-types level (no scarce hotkey consumed — `h` is left-navigation).
- **Persisted per cluster context** (and per named union set) in `~/.local/state/lfk/hidden_types.yaml`, surviving restarts.
- Hidden types **disappear** from the sidebar by default; **`Shift+H`** reveals them **dimmed** (like ignored security findings) so they can be un-hidden from the same row.
- Works for any type — Ingresses, LimitRanges, `_helm/releases`, CRDs, etc.

## Implementation

| Area | File |
|---|---|
| Per-context/union-set persistence (mirrors `pinned.go`) | `internal/app/hidden_types.go` |
| Action-menu hide/show handler (mirrors `handleKeyPinGroup`) | `internal/app/update_actions_hidden.go` |
| Menu wiring + guard-free dispatch | `internal/app/update_actions.go` |
| `model.HiddenTypes` var + `Item.Hidden` field | `internal/model/types.go` |
| Sidebar filter/mark (`markHiddenItems`) | `internal/model/sidebar.go` |
| Scope refresh wired into `applyPinnedTypes` | `internal/app/app_helpers.go` |
| Dimmed rendering (`itemRowStyle`) | `internal/ui/explorer.go` |
| Reveal text, in-app help, startup tip, README, keybindings & config-reference docs | several |

Also fixes a latent slice-aliasing pattern in `togglePinnedTypeIn` (copy before in-place removal), matching the new `toggleHiddenTypeIn`.

## Test plan

- [x] `go build ./...`
- [x] `go test -race ./...` — full suite green
- [x] Unit tests: persistence round-trip, per-context isolation, union-set scope, sidebar filter/mark, pinned+hidden precedence
- [x] Handler tests: hide/show toggle, reveal-on dimmed-stay-and-unhide, action-menu label reflects state
- [x] `go-reviewer` pass; all findings addressed
- [x] Manual TUI check against a live cluster (not run locally)

## Notes

- A type that is both pinned and hidden: **hidden wins** (removed when reveal off; dimmed in the Pinned section when revealed). The status message calls out "removed from Pinned" when hiding a pinned type.


janosmiko:
@coderabbitai review

Issue #339, opened 2026-06-01
[bug]: in list mode, if the sort field is hidden the sort is lost


fallenleavesgocrunch:
### Describe the bug

If you sort by column X in full screen mode of a resource list, when you go back to non-full screen mode that column might not be visible, at which point the sorting is lost.

### Steps to reproduce

1. Go full screen on a resource list
2. Sort by a column that's off to the right
3. Leave full screen mode

### Expected behavior

The list stays sorted the way I left it irrespective of whether the sort column is currently visible

### Actual behavior

With the column not currently being shown it goes to a random unsorted order

### lfk version

0.13.2

### Installation source

Homebrew

### Operating system

macOS 26.5

### Terminal emulator

WezTerm

### Relevant config

```yaml

```

### Relevant log output

```shell

```

### Pre-flight checks

- [x] I searched existing issues and did not find a duplicate.
- [x] I redacted any secrets, tokens, or personal data from this report.

fallenleavesgocrunch:
* When the column isn't in the narrower non-fullscreen mode, it sorts by name

fallenleavesgocrunch:
Works like a charm.

PR #356, opened 2026-06-03
feat(ui): make Name a configurable resource-list column

### LINKED ISSUE #354: [feat]: resource list column configuration
### Problem / motivation

I've noticed there's no real rhyme or reason as to which columns come in what order at the moment.

### Proposed solution

Now that you have the x-menu for resource types that let you pin/hide, it'd be neat if we could, in there, also re-order and hide/show columns for that resource type.  I'd love to be able to order by chores so I get catalog,task,name and hide namespace. That level of customisation would be killer.

### Alternatives considered

_No response_

### Area

Resource list / preview

### Additional context

_No response_

### Pre-flight checks

janosmiko:
Closes #354.

## Summary

Makes the resource-list **Name** column a first-class, configurable column in the column-toggle overlay (`,`). Previously Name was hardcoded as always-first and never listed in the overlay. Now it can be **reordered** (`J`/`K`) and **hidden** (`Space`) like any other built-in column — e.g. arrange `Catalog, Task, Name` and hide `Namespace`.

When visible, Name still renders first by default, absorbs the row's leftover width, and compresses (truncates) when the other columns leave it little room.

## Changes

| Area | Change |
|---|---|
| `orderedColumnKeys` | `"Name"` is now a member of the column order. Backward-compat: saved orders predating this feature (no `"Name"` entry) keep Name pinned first; orders that explicitly list `"Name"` honor that position. |
| Row formatters | `formatTableRowOrdered` / `formatTableRowStyledOrdered` render the Name cell at its position in `order`; `"Name"` presence in the order is the single source of truth (no implicit prepend). |
| `RenderTable` | Computes `hasName` from `ActiveHiddenBuiltinColumns["Name"]`, sets `ActiveNameHidden`, unified header/layout/sortable loops. |
| `collectExtraColumns` | Drops the NAME width reservation when hidden so extras reclaim the freed space. |
| Overlay (`update_column_toggle.go`) | Lists Name; persists through the existing `columnOrder` / `hiddenBuiltinColumns` maps — no new plumbing. |
| Sort reset | Falls back to the first visible column when Name is hidden, mirroring the existing hidden-column guard in sort Next/Prev (issue #339). |
| Docs | `docs/keybindings.md` updated; removed the stale "Name is fixed / never listed" note. |

## Test plan

- [x] New unit/integration tests: `explorer_name_column_test.go`, `update_column_toggle_name_test.go`, `update_keys_sort_reset_test.go` (reorder, hide, default-first backward-compat, sort-reset coherence)
- [x] Updated tests that encoded the old "Name is mandatory" contract
- [x] `go build ./...`
- [x] `go test -race ./...` (no failures or data races)
- [x] `golangci-lint run ./internal/...` (0 issues)
- [x] `gofumpt` clean

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@ Contributions are welcome! Here is how to get started.
 
 ## Prerequisites
 
-- Go 1.26.3 or later
+- Go 1.26.4 or later
 - Access to a Kubernetes cluster (for testing)
 - `kubectl` configured and working
 - `golangci-lint`
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build stage
-FROM golang:1.26.3-alpine@sha256:91eda9776261207ea25fd06b5b7fed8d397dd2c0a283e77f2ab6e91bfa71079d AS builder
+FROM golang:1.26.4-alpine@sha256:f23e8b227fb4493eabe03bede4d5a32d04092da71962f1fb79b5f7d1e6c2a17f AS builder
 
 RUN apk add --no-cache git
 
--- a/docs/keybindings.md
+++ b/docs/keybindings.md
@@ -469,7 +469,7 @@ Press `V` on a resource (or open the Events list and press `Enter` on an event) 
 ## Column Toggle Overlay
 
 Press `,` in the resource list to open the column toggle overlay. It lists
-every toggleable column for the current kind — both built-ins (Namespace,
+every toggleable column for the current kind — built-ins (Name, Namespace,
 Ready, Restarts, Status, Age) and extras from the resource's
 `additionalPrinterColumns`.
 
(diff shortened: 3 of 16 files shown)


janosmiko:
Addressed the CodeRabbit review in a715d43f. Each finding was verified against source before acting:

**Fixed**
- **`mergeColumnToggleEntries` pinned-first Name (Major)** — real bug. The overlay applied a legacy saved order verbatim and appended `Name` later, so reopening the overlay on an order like `["Age","IP","Namespace"]` and pressing Enter relocated `Name` away from first, diverging from `orderedColumnKeys`. Now pins `Name` first when the saved order omits it (mirrors the renderer). Added `TestMergeColumnToggleEntries_LegacyOrderPinsNameFirst` / `_ExplicitNameOrderHonored`, and updated `TestCovOpenColumnToggleRespectsSavedOrder` which had encoded the old behavior.
- **Test isolation of `ActiveHiddenBuiltinColumns` (Minor)** — added save/restore + `nil` in `TestRenderTable_NameReorderedHeader`.
- **Reset test assertion (Nitpick)** — now asserts the exact fallback column (`Namespace`) and `sortAscending == true`.
- **Redundant map in Name-hidden test** — removed (it was dead code, not a compile error — `go vet` was clean — but worth dropping).

**Skipped, with reason**
- **Recompute cached `ExtraCols` when Name visibility changes (Major, outside-diff)** — not a real issue. `TableRenderer.Render` includes `serializeBoolSet(ActiveHiddenBuiltinColumns)` in `tableFingerprint` (`explorer_table_renderer.go:75`); any change to Name visibility changes the fingerprint, which resets `r.layout` (`Computed=false`) and forces a `collectExtraColumns` recompute with the correct `ActiveNameHidden` budget. The freed budget is reclaimed on the next render.

All green locally: `go test -race ./...`, `golangci-lint`, `gofumpt`.

PR #359, opened 2026-06-03
feat: persist resource-list sort order and column layout across restarts

### LINKED ISSUE #353: [bug]: sort order on resource lists is not persisted
### Describe the bug

You can sort the resource list and when you start lfk later it'll be back to default sort

### Steps to reproduce

1. start lfk
2. sort a list
3. quit
4. start lfk
5. check the list sort order

### Expected behavior

~/.local/state/lfk should persist the desired sort column and order, only throwing it away if the CRD no longer has that as a field

### Actual behavior

sort column and order per resource type is not preserved across app restarts

### lfk version

0.13.4

### Installation source

Homebrew

### Operating system

macOS 26.5

### Terminal emulator

WezTerm

### kubectl version

_No response_

### Kubernetes server version

_No response_

### Relevant config

```yaml

```

### Relevant log output

```shell

```

### Additional context

_No response_

### Pre-flight checks

janosmiko:
## Summary

Persists resource-list view preferences across app restarts, so quitting and reopening lfk keeps how you set up each list. Closes #353 (sort order) and extends the same persistence to column visibility/order from the `,` toggle overlay. Mirrors the existing `session.yaml` / `hidden_types.yaml` state-file pattern.

### Sort order (#353) — `~/.local/state/lfk/sort_memory.yaml`
- Persist the existing per-kind sort memory (keyed by context + GVR).
- Per-key merge/delete so a sort in one tab never clobbers another tab's saved sort (sortMemory is per-tab; the file is shared).
- Load at startup into the model + `tab[0]`; seed session-restored tabs from disk (the actual restart path).

### Column layout — `~/.local/state/lfk/column_prefs.yaml`
- Persist the column-toggle overlay's committed layout: visible extras, hidden built-ins, and display order (keyed by context + Kind).
- **Commit-only**: written on `Enter` (commit) and `R` (reset), never on live-apply — so an `Esc` that reverts edits via the snapshot leaves the on-disk layout untouched.
- These three maps are **Model-global** (scoped per context+kind, shared across tabs) rather than per-tab — the natural scope for a "how I view this kind in this cluster" preference, and it keeps persistence to a single source of truth.
- `visible_extras` round-trips a non-nil empty slice ("user configured no extras") distinctly from absent (auto-detect).

### Robustness
A remembered sort column or layout that no longer exists degrades gracefully (sort falls back to the Name tiebreaker; unknown columns are simply not rendered) — a removed CRD field never breaks the list.

## Docs
- `docs/config-reference.md` — two new state-file table entries.
- `docs/keybindings.md` — corrected the stale "session-only" / "not persisted to disk" notes for both sort and column toggle.

## Test plan
- [x] New unit + handler-level tests: persistence round-trip, corrupt/missing files, per-key merge, reset-removal, commit-only invariant, session-restore seeding, nil-vs-empty round-trip.
- [x] Full suite passes with `-race`.
- [x] `gofumpt` clean; `golangci-lint run ./...` -> 0 issues.
- [x] Two Go-reviewer passes; all HIGH findings fixed.
- [ ] Manual: sort + reorder/hide columns, quit, reopen -> both preserved; `R` reset and `-` sort-reset survive restart.

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -917,6 +917,7 @@ The application follows the [XDG Base Directory Specification](https://specifica
 | `$XDG_CONFIG_HOME/lfk/config.yaml` | Main configuration file (default: `~/.config/lfk/config.yaml`) |
 | `$XDG_STATE_HOME/lfk/bookmarks.yaml` | Saved bookmarks (default: `~/.local/state/lfk/bookmarks.yaml`) |
 | `$XDG_STATE_HOME/lfk/session.yaml` | Last session state, auto-managed (default: `~/.local/state/lfk/session.yaml`) |
+| `$XDG_STATE_HOME/lfk/sort_memory.yaml` | Per-context and per-resource-kind sort column/order, auto-managed (default: `~/.local/state/lfk/sort_memory.yaml`) |
 | `$XDG_STATE_HOME/lfk/pinned.yaml` | Per-context and per-union-set pinned CRD groups, managed via `p` key (default: `~/.local/state/lfk/pinned.yaml`) |
 | `$XDG_STATE_HOME/lfk/hidden_types.yaml` | Per-context and per-union-set hidden resource types, managed via the action menu (`x`) at the resource types level (default: `~/.local/state/lfk/hidden_types.yaml`) |
 | `~/.local/share/lfk/lfk.log` (default) | Application log file (configurable via `log_path`) |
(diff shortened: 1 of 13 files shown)


janosmiko:
Thanks @coderabbitai — addressed in 750ee5e7:

**Fixed (both actionable + 2 nitpicks):**
- Cleared `LFK_STATE_DIR` in the XDG/home subtests of `TestSortMemoryFilePath` (and the matching `TestColumnPrefsFilePath`) so a stray `LFK_STATE_DIR` can't break the assertions.
- Dropped the redundant `os.MkdirAll` on `t.TempDir()` in the corrupt-file tests.
- Renamed `TestLoadSortMemorySkipsMalformedKeys` → `TestLoadSortMemoryRebuildsValidKeys`; it feeds well-formed YAML, so it validates reconstruction. The skip-malformed path is on the save side and is already covered by `TestSortMemoryToStateSkipsMalformedKeys`.

**Skipped (with reason):**
- Load `sort_memory.yaml` once for multi-tab restore (`update_bookmarks_session.go:198`): keeping the per-tab `loadSortMemory()` for simplicity — it re-reads a tiny file only during session restore, bounded by tab count (typically 1-3), so the N extra reads are negligible and not on any hot path. Each tab needs its own map copy anyway, which the current code gives for free.

PR #382, opened 2026-06-08
fix: route mouse wheel to the pane under the pointer in Object Explorer and log viewer

### LINKED ISSUE #379: [bug]: object explorer panes don't honor mouse scrolling
### Describe the bug

when you shift+o on a resource, you get tree navigation in the middle and details on the right; moving the mouse to that right pane and doing a scroll action will scroll the middle list, not the details.

### Steps to reproduce

described in the bug description.

### Expected behavior

right pane should scroll because the mouse is over it

### Actual behavior

middle pane scrolles

### lfk version

0.13.9

### Installation source

Homebrew

### Operating system

macOS 26.5

### Terminal emulator

WezTerm

### kubectl version

_No response_

### Kubernetes server version

_No response_

### Relevant config

```yaml

```

### Relevant log output

```shell

```

### Additional context

_No response_

### Pre-flight checks

janosmiko:
## Summary

Fixes #379. In the Object Explorer (`O`), scrolling the mouse wheel over the right YAML preview pane wrongly moved the middle tree cursor instead of scrolling the preview. The Object Explorer fell into the generic `isViewerMode` wheel path, which synthesizes `j`/`k` keypresses regardless of pointer position. The main explorer already solved this with per-pane wheel routing; this wires the Object Explorer (and the log viewer) into the same pattern.

While verifying other multi-pane views, the **log viewer** turned out to have the identical bug: its wheel always scrolled the main log stream, even when the pointer was over the structured **preview side panel**. Fixed here too for consistency.

## Changes

- **Object Explorer** (`handleObjectExplorerWheel`): wheel over the right (preview) pane scrolls the YAML preview; over the left/middle panes it moves the tree cursor. `objectExplorerRightPaneStart` mirrors the column math in `RenderObjectExplorerView`.
- **Log viewer** (`handleLogsMouse` + `scrollLogPreviewWheel`): wheel over the preview side panel scrolls the structured preview (reusing the `J`/`K` handlers so clamping stays in one place); the log stream scrolls otherwise.
- Removed `modeObjectExplorer` from `isViewerMode` now that it has native per-pane handling.
- Extracted the log mouse handling into `handleLogsMouse` to keep `handleMouse` under the gocyclo cap.
- Documented the per-pane wheel behavior in `docs/keybindings.md`.

## Verification sweep

| View | Wheel routing |
|---|---|
| Main explorer | Already per-pane (unchanged) |
| Object Explorer | Fixed |
| Log viewer preview panel | Fixed |
| Explain / "API Explorer" | No bug — DESCRIPTION pane has no independent scroll |
| Centered list overlays | Correct — single list |
| Fullscreen/custom overlays (Can-I, Who-Can, NetworkPolicy) | Wheel swallowed by design (keyboard-only) |

## Test plan

- [x] `go build ./...`
- [x] `go test -race ./internal/app/ ./internal/ui/`
- [x] `golangci-lint run ./...` (0 issues)
- [x] New tests: `update_mouse_objectexplorer_test.go` (right pane scrolls preview / clamps at top; middle pane moves cursor & resets preview scroll) and `update_mouse_logpreview_test.go` (preview pane scrolls preview / clamps; log pane scrolls log; hidden preview falls back to log)
- [ ] Manual: `O` on a resource with a nested status, scroll wheel over the YAML preview → preview scrolls; over the list → cursor moves. In the log viewer with `P` preview on, scroll over the panel vs the stream.

### DIFF
--- a/docs/keybindings.md
+++ b/docs/keybindings.md
@@ -854,6 +854,7 @@ or `Esc` to cancel.
 | Wheel up/down inside a centered overlay | Scroll the list cursor (same as `j` / `k` / arrow keys) |
 | Scroll wheel over middle pane | Move the row selection up/down |
 | Scroll wheel over right pane | Scroll the preview under the pointer |
+| Scroll wheel in the Object Explorer / log viewer | Same per-pane routing — over the preview pane it scrolls the preview, over the list it moves the cursor |
 | `Ctrl+Option+Y` | Toggle mouse capture — release it to select text where `Shift+Drag` doesn't work, press again to re-enable |
 | Shift+Drag | Select text (host terminal) |
 | Shift+Option+Drag (macOS) / Alt+Drag (Linux, Windows) | Block-select text inside the embedded PTY |
--- a/internal/app/update_mouse.go
+++ b/internal/app/update_mouse.go
@@ -62,23 +62,21 @@ func (m Model) handleMouse(msg tea.MouseMsg) (tea.Model, tea.Cmd) {
 
 	// Handle mouse scroll in log viewer mode.
 	if m.mode == modeLogs {
+		return m.handleLogsMouse(msg)
+	}
+
+	// Object Explorer wheel: route by the pointer column. Over the right
+	// (preview) pane the wheel scrolls the YAML preview; over the left and
+	// middle panes it moves the tree cursor — mirroring the main explorer's
+	// per-pane wheel routing (#379). Non-wheel mouse falls through so
+	// tab-bar clicks keep working.
+	if m.mode == modeObjectExplorer {
 		switch msg.Button {
 		case tea.MouseButtonWheelUp:
-			m.logView.follow = false
-			if m.logView.scroll > 0 {
-				m.logView.scroll -= 3
-				if m.logView.scroll < 0 {
-					m.logView.scroll = 0
-				}
-			}
-			cmd := m.maybeLoadMoreHistory()
-			return m, cmd
+			return m.handleObjectExplorerWheel(msg.X, -1)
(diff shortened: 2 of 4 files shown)


janosmiko:
Thanks @coderabbitai — addressed in 43e6fdf6:

**Fixed**
- **Overflowing log fixture (actionable):** `logPreviewWheelModel` now uses a 12-line fixture (overflows the height-10 viewport, `logContentHeight = height-5 = 5`), and `TestLogWheel_LogPaneScrollsLog` / `TestLogWheel_PreviewHiddenScrollsLog` now assert `logView.scroll > 0` so the log-pane routing actually proves the log scrolls.
- **Middle-pane x coordinate:** switched to `x := 0` so the test no longer assumes a ≥10-column left/middle band.

**Skipped (with reason)**
- **`basePush80Model()` + `stripANSI(view)` for the log tests:** these are wheel-*routing* unit tests asserting state transitions, and they mirror the existing sibling pattern in `commands_logs_preview_test.go` (direct `Model` construction). A `stripANSI(view)` assertion exercises rendering, which is orthogonal to routing and already covered elsewhere. The Object Explorer tests already build on `basePush80Model()` via `objectExplorerModel(t)`.
- **Refactor inline log-stream scroll into a shared helper:** that block is pre-existing code moved verbatim into `handleLogsMouse`. There is no existing key handler that does "scroll log by N + `follow=false` + `maybeLoadMoreHistory`" (`j`/`k` move the cursor, not the scroll offset), so extracting a single-caller helper would add indirection and risk behavior drift without reuse benefit. Kept surgical.

PR #404, opened 2026-06-10
fix: pin the affected-resources table while finding-group details scroll

janosmiko:
## Summary

- Fixes the scroll bug in the security finding-group preview (reproducible in Advisor, Falco, Heuristic — any source): shift+j / mouse wheel scrolled the affected-resources table and the description together as one buffer, pushing the table out of view.
- Finding groups now go through the existing split-preview machinery (`hasSplitPreview`) — the same path Deployment > Pods uses — so the affected-resources table is a pinned header and `previewScroll` moves only the group details below the separator. The bespoke `renderSecurityGroupSplitPreview` (which assembled both into one scrollable string) is removed.
- Scroll clamping and the memoized line measurement follow automatically since all three consumers (`clampPreviewScroll`, `measureScrollableLines`, `renderRightColumn`) share `hasSplitPreview`. The memo key now also includes the selected item name so the scroll bound recomputes when switching between two groups with otherwise identical layout dimensions.

## Test plan

- [x] New TUI tests: pinned table survives `previewScroll > 0` while the details title scrolls out; split gates on affected rows being loaded; clamp bounds against the details portion only
- [x] `go test -race ./...` green, `golangci-lint` 0 issues

### DIFF
--- a/internal/app/view_right.go
+++ b/internal/app/view_right.go
@@ -122,6 +122,10 @@ type previewMeasureKey struct {
 	yamlLen     int
 	dashLen     int
 	eventsLen   int
+	// selName invalidates the memo when the middle selection changes while
+	// every layout dimension stays equal — e.g. two security finding groups
+	// with the same affected count but different description lengths.
+	selName string
 }
 
 // measureScrollableLines returns the line count of the scrollable right-pane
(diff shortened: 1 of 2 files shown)


janosmiko:
CodeRabbit's nitpick addressed in 98814e3: added `TestMeasureScrollableLinesRecomputesOnSelectionChange`, which switches between two finding groups with identical layout dimensions but different description lengths and asserts the scrollable line count recomputes. Verified by mutation — the test fails with the `selName` assignment disabled.

PR #405, opened 2026-06-10
feat(preview): live-log preview in the explorer right pane

janosmiko:
## Summary

Adds a per-pod **live-log preview** to the Object Explorer's right pane, toggled with `shift+l`. The fullscreen log viewer moves to `ctrl+l`. When a single Pod is selected the pane tails its logs in real time, isolated from the fullscreen viewer's buffer so the two never clobber each other.

### Streaming
- Dedicated `kubectl logs -f --all-containers` stream per selected pod, with a single-reader guard across model copies.
- Debounced pod switch; auto-reconnect across init-container transitions (mirrors the fullscreen viewer), with no endless re-stream on finished pods (`--tail=0` on reconnect, bounded retry).
- Fresh open fills the pane (initial `--tail` = visible height); lazy older-history fetch on scroll-up prepends the previous batch.
- Per-pod LRU buffer cache: toggling off/on or navigating away and back restores the buffer instantly and resumes new-lines-only.

### Rendering
- Relative-time + message laid out as a table with hanging-indent wrap; pod prefix and absolute timestamp stripped. Lines highlighted and wrapped like the fullscreen viewer.
- Plain dim header consistent with the DETAILS pane.

### Input
- `J`/`K` scroll, mouse-wheel scroll, bottom-anchored follow.
- **Scroll-anchored follow:** while scrolled back, newly arrived lines hold position instead of yanking the view to the newest line; auto-follow resumes only when scrolled to the bottom (matches the fullscreen viewer).
- Configurable keybindings (toggle-preview / viewer keys are swappable).
- `log_viewer.preview_live` startup default.

### Defaults
- Fullscreen viewer and preview tail default to 100 lines.

### Shared helpers
Three stateless helpers — `kubectlPodLogArgs`, `mergeOlderLogLines`, `capLines` — are reused by both the fullscreen viewer and the preview; each surface keeps its own buffer and scroll state.

### Docs
README, `docs/keybindings.md`, in-app help, `docs/config.schema.json`, and `docs/config-example.yaml`/`config-reference.md` updated.

> Note: this branch was based on `main` before the audit PR (#401) landed, so the merge is a 3-way merge against current `main`.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/app/ ./internal/ui/`
- [x] `go test -race ./internal/app/`
- [x] Manual: toggle preview on a Pod, scroll back, confirm view holds while new lines stream; scroll to bottom, confirm follow resumes
- [x] Manual: finished/Completed pod does not endlessly re-stream
- [x] Manual: navigate away and back to a pod restores the buffer


shoce:
is there any way to have the previous hotkey L to open the logs viewer? i got very used to it :)

shoce:
also now on my macbook there is only one side of keyboard to do CTRL+L to see logs, before i could do it with one hand with the right SHIFT+L 😢

PR #429, opened 2026-06-14
fix(scheduler): wire unscheduled K8s reads through the task scheduler

janosmiko:
## Summary

An audit of scheduler wiring found K8s operations dispatching their API calls in raw `tea.Cmd` closures (or via `trackBgTask`, which only drives the bgtask indicator) — bypassing the scheduler's bounded worker pool, priority lanes, coalescing, and gen-based cancellation. Every such closure is an unbounded, unprioritized goroutine competing with the pool.

This routes both **reads** and **in-process write mutations** through `scheduleK8sCall`. Two commits:

### 1. Reads → `scheduleK8sCall`

| Loader(s) | Kind / Priority | Notes |
|---|---|---|
| `copyYAMLToClipboard`, `copyYAMLForScope`, `exportResourceToFile`, `loadDiff` | `KindYAMLFetch` / High | Distinct `Copy/Export/Diff` task names so a copy never coalesces onto an in-flight `YAML:` preview fetch (which would deliver the copy a nil result and drop it). |
| `loadContainersForAction`, `loadContainersForLogFilter` | `KindContainers` / High | Per-consumer names — distinct result-msg types. |
| `loadSecretData`, `loadConfigMapData`, `loadLabelData`, `loadRevisions`, `loadPodsForAction`, `detectExecPodOSCmd` | `KindResourceList` / High | `detectExecPodOSCmd` keeps its 5s budget, derived from the scheduler ctx. |
| `loadRightsizing` | `KindMetrics` / Low | Heavy enrichment; yields to foreground. |
| `watchArgoWorkflow`, `loadAutoSyncConfig` | `KindResourceList` / High | Moved off `trackBgTask`. |

### 2. In-process write mutations → `scheduleK8sCall(PriorityCritical, KindMutation)`

30 single in-process client-go writes across gitops (Argo / Flux / cert-manager / KEDA / ExternalSecrets), `exec_misc` (delete / scale / resize / rollback / triggerCronJob), karpenter, knative, and the dataops saves (which were raw closures — not even tracked). `KindMutation` **NeverCoalesces** and is never gen-superseded, so there is no coalescing hazard; the only sentinel a mutation can hit is `ErrContextSwitched`. `:tasks` visibility is preserved (the worker calls `startWithPriority` for every submission).

## Deliberately kept on `trackBgTask` (with reason)

- **Subprocess mutations** (kubectl / helm shell-outs): `forceDeleteResource`, `removeFinalizers`, `restartResource` (rollout restart), node cordon/drain, `kubectlNodeCmdFromClaim`. Not in-process client-go calls; the scheduler ctx/30s budget don't fit.
- **Bulk-loop mutations** (N calls per task → 30s-clamp risk): `bulkSyncArgoApps`, `bulkRefreshArgoApps`, `bulkRemoveFinalizer`. `commands_bulk.go` keeps its `StartCancellable` + progress-`Total` pattern.
- **`cmdLoadOrphans`** — bespoke namespace-scoped `context.WithCancel` + inflight dedup; wiring would drop that cancellation trigger.
- **`loadCrashInvestigation`** — needs a 60s budget that `scheduleK8sCall` would clamp to the 30s default.
- **`loadContexts` / `ReloadKubeconfig`** — read in-memory / on-disk kubeconfig, not the cluster API.

## Why it matters

Off-scheduler K8s calls lose **bounded concurrency** (each was an unbounded goroutine), **priority** (reads yield to foreground; mutations run Critical), **coalescing** (reads dedup duplicate in-flight fetches), and **gen-based cancellation**. They also now appear in the `:tasks` overlay.

## Test plan

- [x] `TestK8sReadsAreScheduled` (15) + `TestK8sMutationsAreScheduled` (13) assert each op `Submit`s synchronously to the scheduler queue at Cmd-construction time. TDD: red against the raw closures / `trackBgTask`, green after wiring.
- [x] Existing roundtrip tests that execute these cmds use a new `execScheduled` helper that starts the worker pool (a worker-less scheduler would otherwise block); models lacking a scheduler get one injected locally.
- [x] `go test -race ./...` — all green, no hangs
- [x] `golangci-lint run` — 0 issues
- [x] `go build ./...`, gofmt/gofumpt clean

### DIFF
--- a/internal/app/commands.go
+++ b/internal/app/commands.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/atotto/clipboard"
 	tea "github.com/charmbracelet/bubbletea"
+	"github.com/janosmiko/lfk/internal/app/scheduler"
 	"github.com/janosmiko/lfk/internal/logger"
 	"github.com/janosmiko/lfk/internal/model"
 	"github.com/janosmiko/lfk/internal/ui"
@@ -186,10 +187,16 @@ func (m Model) loadPodsForAction() tea.Cmd {
 	ns := m.actionNamespace()
 	kind := m.actionCtx.kind
 	name := m.actionCtx.name
-	return func() tea.Msg {
-		items, err := m.client.GetOwnedResources(context.Background(), kctx, ns, kind, name)
-		return podSelectMsg{items: items, err: err}
-	}
+	return m.scheduleK8sCall(
+		scheduler.PriorityHigh,
+		scheduler.KindResourceList,
+		"List pods: "+kind+"/"+name,
+		bgtaskTarget(kctx, ns),
+		func(ctx context.Context) tea.Msg {
+			items, err := m.client.GetOwnedResources(ctx, kctx, ns, kind, name)
+			return podSelectMsg{items: items, err: err}
+		},
+	)
 }
 
 // loadPodsForLogAction fetches pods matching the parent resource's selector using kubectl.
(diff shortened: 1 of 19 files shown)


janosmiko:
Addressed in 124eb3d8 — `GetWorkflowStatus` now takes `context.Context` as its first parameter and passes it to the dynamic `Get`, so the scheduler can cancel/time out `watchArgoWorkflow`'s fetch (consistent with `GetAutoSyncConfig`). Updated both callers. Build, `-race` tests, and lint all green.

PR #443, opened 2026-06-17
feat: force-delete and evict replicas for Longhorn nodes

janosmiko:
## Summary

Adds Longhorn-aware node operations to the `longhorn.io/nodes` ("Longhorn Nodes") resource type.

**Why:** deleting a Longhorn node fails with `admission webhook "validator.longhorn.io" denied the request: could not delete node ... node schedulable true`. The webhook rejects the DELETE itself when the node is still schedulable — and `kubectl delete --grace-period=0 --force` / finalizer removal do **not** bypass validating webhooks. The only real fix is to make the node satisfy the webhook (disable scheduling), then delete.

## What's included

- **Force Delete** (`X`): patches `spec.allowScheduling=false`, then deletes via the dynamic client with the exact GVR (avoids the core-vs-longhorn `nodes` ambiguity). The webhook still protects nodes that hold data, surfacing a clear error instead of risking data loss.
- **Evict Replicas** (`e`) / **Cancel Eviction** (`C`): toggle `spec.evictionRequested` (+ disable scheduling). Longhorn rebuilds each replica elsewhere before removing it, so eviction is data-safe. Use this to drain a node before force-deleting.
- **REPLICAS column** on the Longhorn Nodes list: one extra LIST per refresh grouped by `spec.nodeID` (no N+1) — see at a glance whether a node still holds data.
- **Dedicated Longhorn-node action menu**: the list previously showed the core-node menu (Cordon/Drain/Taint) since both share Kind `"Node"`; detection is group-aware via `IsLonghornNode`.
- **Confirm-overlay wrapping fix**: word-wrap on whitespace (keeping hyphenated names intact), hard-split only over-long tokens, so long resource names no longer shatter mid-word. Also improves the plain Delete / Force Delete confirms.

## Typical workflow (orphaned autoscaled node)

Open Longhorn Nodes → check `REPLICAS` is `0` → `X` Force Delete. If `REPLICAS > 0`, `e` Evict Replicas first, wait for it to drain, then force delete.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race ./...`
- [x] Unit tests: group-aware detection, dedicated menu, force-delete routing (and no leak onto core nodes), evict/cancel confirm flow, read-only gating, REPLICAS column counting by `spec.nodeID`, confirm-text wrapping
- [ ] Manual: force-delete a real stuck Longhorn node; evict replicas off a node with data

### DIFF
--- a/docs/keybindings.md
+++ b/docs/keybindings.md
@@ -971,6 +971,11 @@ The action menu (`x` key) shows context-specific actions based on the resource t
 ### Node Actions
 `c` Cordon, `u` Uncordon, `n` Drain, `t` Taint, `T` Untaint, `s` Shell, `v` Describe, `E` Edit, `b` Debug Pod, `V` Events
 
+### Longhorn Node Actions
+`e` Evict Replicas, `C` Cancel Eviction, `v` Describe, `E` Edit, `D` Delete, `X` Force Delete, `V` Events, `P` Permissions
+
+The Longhorn Nodes list shows a `REPLICAS` column with the count of replicas scheduled on each node. Force Delete disables scheduling, then deletes (the `validator.longhorn.io` webhook rejects deleting a still-schedulable node). Evict Replicas disables scheduling and sets `spec.evictionRequested`, so Longhorn rebuilds each replica on another node before removing it.
+
 ### Job Actions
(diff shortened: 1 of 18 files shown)


janosmiko:
@coderabbitai review

janosmiko:
@coderabbitai full review

PR #448, opened 2026-06-17
feat: Port Forward & Open action for Services (Ctrl+O)

janosmiko:
## Summary
A Service has no directly openable URL, so `Ctrl+O` did nothing useful on one. This adds a **"Port Forward & Open"** flow: start a port forward to the Service, then open `http://localhost:<local-port>` in the browser once the local port resolves.

Exposed two ways:
- New **"Port Forward & Open"** Service action-menu item (key `O`).
- **`Ctrl+O`** on a Service routes to the same chained action.

## How it works
- `pfOpenInBrowserAfterStart` intent flag on `Model`, set by the action and consumed when the port resolves — immediately on start for an explicit local port, or later in `updatePortForwardUpdate` for a random (`0`) port.
- `consumePortForwardBrowserOpen` centralizes the open-and-clear logic (opens only on a resolved, non-`0` port).
- Intent is cleared on start error and on overlay cancel / invalid input, so it never leaks into an unrelated future forward.
- Read-only gated (added to `mutatingActions`) and union-policy scoped to `Service`.

## Notes for reviewers
This PR is **stacked on `feat/open-browser-port-forwards` (#445)** because both touch `handleExplorerActionKeyOpenBrowser`. Base is set to that branch so the diff shows only the Service additions. Once #445 merges to `main`, I'll retarget this PR to `main`.

## Test plan
- [x] `go build ./...`
- [x] New tests: `internal/model/actions_port_forward_open_test.go` (action present/absent, key `O`), `internal/app/port_forward_open_test.go` (intent helper, dispatch, read-only block, start-known/deferred/error, Ctrl+O on Service, overlay-cancel cleanup) — TDD
- [x] `internal/app/readonly_union_closure_test.go` updated for the new union policy row
- [x] `go test ./...` (full suite)
- [x] `go vet` + `golangci-lint` clean; 800-line cap on `app.go` respected

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,7 +382,7 @@ All search and filter inputs support three modes, auto-detected from the query s
 - Teleport between levels with `0` / `1` / `2` (clusters / resource types / resources)
 - Sort by any column with `>` / `<`, flip direction with `=`, reset with `-`
 - Check firing Prometheus/Alertmanager alerts with `@` and namespace quotas with `Q`
-- Open an Ingress host - or an active port-forward's localhost URL - in your browser with `Ctrl+O`
+- Open an Ingress host - or an active port-forward's localhost URL - in your browser with `Ctrl+O`; on a Service, `Ctrl+O` starts a port forward and opens it
 - Save the selected resource manifest to a file with `W`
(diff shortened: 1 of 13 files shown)


janosmiko:
@coderabbitai review

PR #449, opened 2026-06-17
fix: scope to default namespace when toggling all-namespaces off with none set

janosmiko:
## Summary
Pressing the all-namespaces key (`A` / `Shift+A`) to turn the mode **OFF** restored the previous namespace selection — but when there was none to return to (the app started in all-namespaces mode, or the namespace had been cleared by a prior navigation), `m.namespace` stayed empty. `effectiveNamespace()` then returns `""`, which the fetch treats as "all namespaces", so the toggle looked like a no-op and the status bar showed `(ns: )`:

```
All namespaces mode OFF (ns: )
All namespaces mode ON
```

## Fix
When toggling OFF with no concrete namespace and no restored single selection, fall back to the context's default namespace (kubeconfig namespace, or `"default"`). A restored single/multi-namespace selection still takes precedence, so existing behavior is unchanged.

Added `defaultNamespaceForContext` helper (nil-client safe for tests).

The namespace selector overlay (`\`) is unaffected: its `A` only turns the mode ON, and Enter always applies a concrete namespace or all-ns — so the empty-on-OFF path is unique to the explorer toggle.

## Test plan
- [x] `TestActionKeyAllNamespacesOffFallsBackToDefault` — reproduces the empty-namespace symptom, asserts fallback to `default` and non-empty `effectiveNamespace()` (TDD, RED first)
- [x] `TestActionKeyAllNamespacesOffKeepsRestoredSelection` — guards that a restored selection is not clobbered by the fallback
- [x] Existing all-namespaces toggle tests still pass
- [x] `go test ./internal/app/`, `go vet`, `golangci-lint` clean

### DIFF
--- a/internal/app/commands_load.go
+++ b/internal/app/commands_load.go
@@ -622,6 +622,16 @@ func (m Model) resolveNamespace() string {
 	return m.namespace
 }
 
+// defaultNamespaceForContext returns the namespace to scope to when none is
+// set: the active context's kubeconfig namespace when a client is available,
+// falling back to "default". Guards against a nil client (test models).
+func (m Model) defaultNamespaceForContext() string {
+	if m.client != nil {
+		return m.client.DefaultNamespace(m.nav.Context)
+	}
+	return "default"
+}
+
 // loadRevisions fetches the revision history for a deployment.
 func (m Model) loadRevisions() tea.Cmd {
 	sel := m.selectedMiddleItem()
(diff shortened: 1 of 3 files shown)


janosmiko:
Addressed in 3ee83603 — the OFF status now builds the namespace hint from `effectiveNamespace()`, so a restored single selection shows `(ns: <name>)` instead of `(ns: )`. Added a test assertion to lock it in. Thanks @coderabbitai.

PR #451, opened 2026-06-17
fix(templates): make workloads + Service/Ingress compose out of the box

janosmiko:
## Problem
Creating the default **Pod + Service + Ingress** templates did not work out of the box: the Service selected nothing. The Service selected `app: my-app`, but no workload template used that label (Pod was `app: my-pod`, Deployment `app: my-deployment`, etc.).

## Fix
Standardize every basic workload on a single generic **`app: my-app`** label so the Service (and the other glue templates) compose with whichever workload the user creates — Pod, Deployment, ReplicaSet, StatefulSet, or DaemonSet.

| Template | Change |
|---|---|
| Pod, Deployment, ReplicaSet, StatefulSet, DaemonSet | app label (+ controller selector & pod-template label) → `my-app` |
| Service | selector → `my-app`; add `app: my-app` label (so ServiceMonitor selects it) |
| NetworkPolicy | podSelector → `my-app` |
| ServiceMonitor | selector → `my-app` |
| ClusterRoleBinding | subject namespace `default` → `NAMESPACE` (matches ServiceAccount + RoleBinding) |

Resource **names** (`my-pod`, `my-deployment`, …) and the **HPA** `scaleTargetRef` (`my-deployment`) are unchanged — only the `app` label is unified. So: Pod/Deployment/… (`app: my-app`, port 80) ← Service (selects `my-app`, 80→80) ← Ingress (`my-service`:80); NetworkPolicy/ServiceMonitor target the same app; HPA→Deployment and the RBAC bindings resolve consistently.

> Note: using one shared `app` label means creating two controllers (e.g. a Deployment and a ReplicaSet) from the templates unedited would give them overlapping selectors. That's an inherent property of a single generic label and acceptable for starter templates — users customize names/labels per app.

## Test plan
- [x] New `TestTemplatesCrossReferencesResolve`: every workload shares `app: my-app`; Service/Ingress/NetworkPolicy/ServiceMonitor/HPA + RBAC bindings resolve to the right targets (TDD)
- [x] Existing `templates_test.go` still passes
- [x] `go build ./...`, `go test ./internal/model/`, `go vet`, `golangci-lint` clean

### Out of scope
`StatefulSet.spec.serviceName: my-statefulset` references a self-named governing headless Service (standard convention) — left as-is.

### DIFF
--- a/internal/model/templates.go
+++ b/internal/model/templates.go
@@ -248,7 +248,8 @@ spec:
   selector:
     app: my-app
   ports:
-    - port: 80
+    - name: http
+      port: 80
       targetPort: 80
       protocol: TCP
 `,
@@ -510,7 +511,7 @@ spec:
     matchLabels:
       app: my-app
   endpoints:
-    - port: metrics
+    - port: http
       interval: 30s
       path: /metrics
 `,
--- a/internal/model/templates.go
+++ b/internal/model/templates.go
@@ -24,7 +24,7 @@ metadata:
   name: my-pod
   namespace: NAMESPACE
   labels:
-    app: my-pod
+    app: my-app
 spec:
   containers:
     - name: nginx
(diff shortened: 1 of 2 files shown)


janosmiko:
Addressed in 326d0d70 — named the Service port `http` and pointed the ServiceMonitor endpoint at it, so the scrape target references an existing Service port. The cross-reference test now asserts the ServiceMonitor endpoint port matches the Service's named port. Thanks @coderabbitai.

PR #453, opened 2026-06-19
feat: Log Top — live log aggregation view (top-style request analytics)

janosmiko:
## Log Top — live log aggregation ("top" for request logs)

Adds a fullscreen **Log Top** view that parses a resource's logs and groups them by request attributes (method, path, host, status, router/service) showing per-group request counts, rate, error counts, and latency percentiles — a `top`/goaccess-style live analytics view over Kubernetes logs.

Launch from the resource action menu (**"Log Top"**, quick-key `T`) on any log-capable resource, or press `T` from the open log viewer. Opens as a snapshot of the loaded buffer and keeps updating live from the stream.

### Formats (auto-detected, overridable with `p`)

| Profile | Source | Notable fields |
|---|---|---|
| `traefik-json` | Traefik JSON access logs | method, path, **host**, status, duration |
| `nginx-combined` | nginx / Apache / Traefik CLF | method, path, status, duration, **router** (Traefik) |
| `ingress-nginx` | NGINX Ingress Controller | method, path, status, duration, **service** (upstream) |
| `envoy` | Istio / Contour / Emissary / Gloo / Gateway API | method, path, status, duration, **host** (`:authority`), service |
| `json` / `logfmt` | structured app logs (+ Envoy/Caddy JSON via aliases) | any present keys |

Detection prefers a structured HTTP profile over the generic `logfmt`/`json` matchers when present, so mixed Traefik output (JSON access logs + console error lines) reliably detects as `traefik-json`.

### Features

- **Multi-dimension columns** — method/path/status/host/router/service; non-grouped dimensions show their value, or `*N` when varied within a group.
- **Metrics** — REQ, REQ/s, %, ERR, and **p50/p95/p99** latency (bounded log-histogram; `n/a` without duration data).
- **Drill-down** — `Enter` descends to the next dimension (marked `▸` in its header); `Esc` returns to the exact prior level. `Tab` cycles which dimension is next.
- **Sort** — `>`/`<` cycle all columns, `=` flips direction (active column shows ↑/↓), `-` resets; matches the explorer's header styling.
- **Filter** (`f`) and **search** (`/` with `n`/`N`) over rows.
- **Column selector** (`,`) — show/hide + reorder columns, identical to the explorer's.
- **Full vim navigation** — `ctrl+d`/`u`/`f`/`b` + shift-arrow/pgup-pgdn aliases, with scroll-following.
- **Per-tab persistence**; **bounded memory** (incremental aggregation, capped distinct sets and parsed cache, fixed-bucket histograms — no per-request retention).

### Implementation

- New pure, dependency-free `internal/logagg` package (parsers + aggregator), wired into a `modeLogTop` viewer in `internal/app` and a renderer in `internal/ui`.
- New config key `log_top_default_profile` (with schema + docs + wiring tests).
- Reuses existing primitives: log stream, `ui.RenderOverlayList`, `renderStyledHeader`/sort-arrow, `VimScrollOff`, `MatchLine`, `TextInput`.

### Testing

- ` 59 files changed, 4890 insertions(+), 59 deletions(-)`
- Unit + table-driven tests for every parser (with real sample lines), the aggregator, percentile math, and detection priority.
- TUI tests via `basePush80Model()` + `stripANSI(view)` for rendering, navigation, sort, filter, search, drill, and the overlays.
- Config forcing-function tests for the new setting.
- Full suite green with `-race`; `golangci-lint` clean.

Docs updated: README, `docs/keybindings.md`, in-app help, config reference/example/schema.


janosmiko:
## CodeRabbit findings — resolution

Thanks for the review. Addressed in `c32464bf` (metrics) and `ea52c3bd` (fixes):

**Fixed**
- Latency `n/a` rows now sort **last** in both directions (new `cmpLatency`, applied to AVG/P50/P95/P99/MAX).
- Envoy host/service extraction now requires the full default-format token count (`len(q) >= 5`) before assigning, so truncated/non-default lines no longer mislabel dimensions.
- nginx Traefik-router regex tightened to the trailing field position, so a referer/UA containing `@` is no longer captured as the router.
- `LogTopDefaultProfile` comments (config_load.go + config.go) now list all 7 profiles.
- Log Top sort help text corrected (cycle column / flip / reset, not "REQ ↔ ERR").
- Metric padding uses `lipgloss.Width` instead of `len`.
- Drill-hint test now asserts the precondition explicitly.
- Vim **count-prefix** support added for Log Top motions (`5j`, `10G`, paging) via `consumeCountPrefix`.

Also added (unrelated to review): avg/max/error-rate/4xx/5xx metrics, and a fix where streaming Traefik logs locked the wrong profile (re-detection over the buffer).

**Intentionally not changed**
- `inputActive()` `modeExplorer` guard — pre-existing behavior (the explorer handles tab keys in its own dispatch); this PR only extracted the existing condition into a helper, so it's not a regression.
- `g` = group-by in Log Top is a deliberate binding (shown in the hint bar); top-of-list is on `home`. Happy to switch to vim `g`/`G` and relocate group-by if preferred.

PR #491, opened 2026-07-04
feat(kubeconfig)!: make KUBECONFIG exclusive by default, with opt-out

qjoly:
## Summary

Make the `KUBECONFIG` environment variable **exclusive**, matching kubectl and k9s semantics.

Previously, lfk read `KUBECONFIG` but always merged `~/.kube/config` and auto-scanned `~/.kube/config.d/` on top of the listed files. That diverged from how kubectl behaves: setting `KUBECONFIG` there replaces the default kubeconfig entirely.

**Imitating kubectl's behavior here is intentional.** Users who set `KUBECONFIG=/path/to/config` expect to see only that config, not also every cluster discovered under `~/.kube/config.d/`. Aligning with the ecosystem tool (kubectl/k9s) is the least surprising behavior.

## Behavior after this change

When `KUBECONFIG` is set:
- only the files it lists are loaded;
- the default `~/.kube/config` is **not** added;
- the default `~/.kube/config.d/` scan is **not** performed;
- directories the user **explicitly** requests (`--kubeconfig-dir`, `KUBECONFIG_DIR`, `kubeconfig_dir`) are still merged, since those are deliberate opt-ins rather than implicit defaults;
- `--kubeconfig` remains the strongest override and bypasses all discovery.

When `KUBECONFIG` is unset, behavior is unchanged (`~/.kube/config` + `config.d/` discovery).

## Changes

- `internal/k8s/client_kubeconfig.go` — `buildKubeconfigPaths` skips the default paths when `KUBECONFIG` is set.
- `internal/k8s/client.go` — updated `NewClient` doc comment.
- `internal/k8s/client_test.go` — new tests locking the exclusive behavior and the explicit-dir merge exception.
- Docs/schema: `README.md`, `docs/usage.md`, `docs/config-reference.md`, `docs/config.schema.json`, `docs/config-example.yaml`.

## Tests

`go build ./...` and the full `internal/k8s` + `internal/ui` test suites pass. Existing `KUBECONFIG` tests were unaffected (dir-discovery tests already set `KUBECONFIG=""`).

---

## Maintainer update (2026-07-07)

Per the discussion below, exclusivity ships **enabled by default** (security-first, kubectl semantics) with an opt-out escape hatch instead of an opt-in flag:

- `kubeconfig_exclusive` config key (bool, default `true`)
- `--kubeconfig-exclusive` CLI flag
- `LFK_KUBECONFIG_EXCLUSIVE` env var
- Precedence: CLI > env > config > default (`true`)

The follow-up commit also folds in the earlier review items: empty `KUBECONFIG` entries are filtered (an all-empty value such as `KUBECONFIG=":"` counts as unset, so default discovery still applies instead of loading zero clusters), the dead `containsPath` guard is removed, the trivially-passing dedup subtest now exercises the merge path, subtest names match the house style, a test pins `--kubeconfig` as the strongest override, and the config key follows the four-part config contract (struct+apply, schema, docs, wiring tests).

## BREAKING CHANGE

When `KUBECONFIG` is set, lfk no longer merges `~/.kube/config` and the default `~/.kube/config.d/` scan — clusters outside `KUBECONFIG` disappear from the cluster list unless explicitly re-added. Restore the old behavior with `kubeconfig_exclusive: false` (or `LFK_KUBECONFIG_EXCLUSIVE=false` / `--kubeconfig-exclusive=false`).

**Release note:** the next release including this PR must bump the minor version (0.x semver: breaking → minor).


janosmiko:
Hi @qjoly ,

Thanks for your pull request. 

I don't really understand the purpose of this. I mean... in my use-case, I constantly have KUBECONFIG env var set in my shells, and that sets my default cluster which is also marked in lfk in the cluster view.

I don't see the additional value of this feature. Could you explain me why would it good to add this to lfk?
I understand the references (kubectl and k9s) but both of these tools are designed to manage single-clusters, while lfk is designed to manage multi-clusters and I think the users expect the current behaviour.

If you really want to add this (for whatever reason), we can add, but gate it with a flag (default to be disabled) and users who wants to hide the rest of the clusters could enable this feature imho.

But I'm open to discussion.

qjoly:
What I had in mind is more about scope.

The way I think about it, KUBECONFIG is how most tools decide which clusters they're allowed to talk to. When I point it at a single file (a prod-only kubeconfig, or one customer), I'm drawing a line for that session. kubectl, k9s and client-go all read it that way, so I've gotten used to relying on it as a boundary.

With the current merge behaviour, lfk still connects to everything under ~/.kube/config and ~/.kube/config.d/, including clusters I meant to leave out. In practice that can fire exec credential plugins for clusters I didn't intend to touch (aws eks get-token, an SSO prompt) just from opening the app, and reach API servers I wasn't planning to hit. That matters more in consulting or MSP setups where each customer's kubeconfig is kept on its own.

So for me it's mostly a convention and safety thing, and it doesn't really cost the multi-cluster workflow anything, since KUBECONFIG takes a colon-separated list and --kubeconfig-dir / KUBECONFIG_DIR still merge.

That said, it's your call and I'm happy with the gated approach. I can put it behind a flag defaulting to the current behaviour, so nothing changes for existing users. Do you have any preferred flag name (or var env) ?

janosmiko:
Hi @qjoly,

Thanks for the detailed explanation - the exec-credential-plugin point is convincing. Happy to take this **behind an opt-in flag**.

Naming: let's go with kubeconfig_exclusive (bool, default `false`) as the config key, with `--kubeconfig-exclusive` as the CLI flag and `LFK_KUBECONFIG_EXCLUSIVE` as the env override (matching the existing LFK_* pattern, e.g. LFK_ICONS).

Heads-up on the config plumbing: new config keys in this repo have a four-partcontract enforced by CI (see the "Configuration Settings" section in CLAUDE.md):

1. Field on configFile in internal/ui/config_load.go + wiring in `internal/ui/config_apply.go`
2. Property in `docs/config.schema.json`
3. Docs: `docs/config-reference.md` table + commented example in `docs/config-example.yaml`
4. Wiring test in `internal/ui/config_wiring_test.go` (TestLoadConfig_AllSettingsWired + add the field to wiringCoveredFields)

`TestConfigSchemaTopLevelInSync` and `TestConfigFile_EveryFieldHasWiringCoverage` will fail until all four are in place, so doing them together saves a review round-trip.

A few smaller things to fold into the same revision:
- Empty `KUBECONFIG` entries: with exclusivity on, a stray `KUBECONFIG=":"` or trailing colon currently produces empty path entries and zero discovery (pre-PR the default fallback still kicked in as a safety net). Please filter empty entries from `filepath.SplitList` output, and add a test for it.
- The `containsPath(paths, defaultPath)` guard is dead now that the block is gated on !kubeconfigSet (CodeRabbit flagged this too) - the dedup pass at the end of `buildKubeconfigPaths` already covers it, so it can just be an append.
- The pre-existing subtest "does not duplicate default path when in KUBECONFIG" passes trivially under the new gating - worth updating or renaming so it still tests something meaningful (probably only relevant in the flag-off path after gating).
- Subtest naming: the existing TestBuildKubeconfigPaths subtests use lowercaseimperative names ("includes...", "dedups..."); please match that style for the new ones.
- Since the PR body claims --kubeconfig stays the strongest override, a small test pinning that would be welcome too.

With the flag defaulting to `false`, nothing changes for existing users and your scoped workflow gets the kubectl semantics. Looking forward to the update!

GitBluub:
hi @janosmiko 👋 

I think setting this as an opt-in is wrong for a few reasons:
- When a user defines the KUBECONFIG env var to set guardrails in a specific shell, I don't think we should not respect them by default. User defined guardrails like this are even more frequent with local agents now.
- this feature as a default isn't against multi-cluster philosophy, I do (and I think it's pretty standard) to want to split personal cluster / work clusters, or groups of work cluster for different clients. If the user specifically set the KUBECONFIG env var as guardrails for his kubectl commands (and other tools in the kube ecosystem) it makes sense that lfk uses it like the rest of the eco-system as a default.
- lfk is clearly designed for multi-cluster but I also find value in mono cluster view compared to k9s, I love a lot of the things bundled in and i regularly use multiple tabs for a single cluster.

Anyway it's your choice, just wanted to chime in :D

janosmiko:
Hi @GitBluub ,

Thanks for adding your opinion on this. As I mentioned above, I agree with the security concern, just wanted to add it behind an opt-in flag to make sure users are not getting surprised by this.
But both of you are right, and I handle security as a top priority for this project, so I'll add this as an opt-out flag instead.

Thanks for bringing this up @qjoly

janosmiko (dismissed):

janosmiko:
@qjoly I've pushed the agreed direction to your branch (thanks for enabling maintainer edits): exclusivity is now the **default** with an opt-out — `kubeconfig_exclusive` config key (default `true`), `--kubeconfig-exclusive` CLI flag, and `LFK_KUBECONFIG_EXCLUSIVE` env var (CLI > env > config). The commit also covers the review checklist from my earlier comment (empty-entry filtering with unset fallback, dead `containsPath` removal, subtest fixes/naming, `--kubeconfig` override pin, and the four-part config contract). PR is retitled with the `!` breaking marker and labeled accordingly. Please take a look — if it matches what you had in mind, this is ready to go.

janosmiko (dismissed):

PR #509, opened 2026-07-08
feat(sessions): named sessions with an active-session model

janosmiko:
## Summary

**Named sessions** = whole multi-tab workspace snapshots you can save, switch between, and manage. lfk always has one **active session** that auto-saves on quit and restores on start, so working in one session never clobbers another (or the default).

- **Session manager**: `C` (rebindable) or `:sessions` — a picker overlay.
- **Commands**: `:session save <name>`, `:session delete <name>`.
- **In the overlay**: `enter` switch, `s` save current workspace as a new name (prompts), `d` delete, `/` filter; a ✓ marks the active session.
- **CLI**: `lfk --session <name>` / `LFK_SESSION` opens a session at startup (created on first save if new; mutually exclusive with `--context`).

## Model

- A session is a `SessionState` (all tabs: context, namespace scope, resource type/name, filter, cursor + active tab).
- The **default** workspace stays in `session.yaml` (no migration) and shows as a synthetic `default` row in the picker — always reachable.
- **Named** sessions live one-file-per-session at `<StateDir>/sessions/<sanitized>.yaml` as `NamedSession{Name, SavedAt, State}`.
- The **active session** is a small `<StateDir>/active_session` pointer (empty = default). Precedence at startup: `--session` / `LFK_SESSION` → persisted pointer → default. `saveCurrentSession` writes to the active session's file; switching saves the one you leave, then rides the existing `pendingSession`/`restoreMultiTabSession` startup restore path.
- Global state (columns, sort, hidden types, pinned, bookmarks, command history, local clusters, port forwards) stays shared across sessions — switching changes *where you look*, not tool preferences.

## Guards

- Same-name save overwrites (update); a different name that sanitizes to the same file is rejected (data-loss guard).
- Save and switch are rejected in union view. The `default` row cannot be deleted.

## Test plan

- [x] Storage: named-session round-trip + collision guard; active-session pointer round-trip; startup resolution precedence; save targets the active session
- [x] Overlay: default row + active marker, switch to named/default (persists active), delete (default rejected), save-as prompt, filter, `C` opens the manager
- [x] `--session` flag + `LFK_SESSION` env
- [x] Keybinding schema + wiring tests; docs/help/hint updated
- [x] `go build`, `go test ./...`, `go test -race ./internal/app`, `golangci-lint run ./...` — all green

Built via spec → plan → subagent-driven for the base feature (with an Opus whole-branch review that caught the collision guard), then the active-session model added on top with inline TDD.


janosmiko:
Addressed the two Major findings from the latest CodeRabbit review in f57bfb28:

1. **`activeSession` not reset on CLI override** — `lfk --session foo --context bar` would overwrite `foo.yaml` with the synthetic single-tab workspace on quit. Fixed: the `HasCLIOverrides()` branch in `NewModel` now clears `activeSession` so auto-save targets the default (`session.yaml`). `--session foo` alone still stays active. Test: `TestNewModel_SessionClearedByCLIOverride`.
2. **Swallowed `loadNamedSession` failure in `switchToSession`** — a failed load switched into an empty workspace under the name (which would then overwrite the file). Fixed: load the target *before* committing the switch; on failure, keep the current session active and the overlay open with an error. Test: `TestSessionsOverlaySwitchLoadFailureAborts`.

The earlier `j/k` hint nit was already resolved in 37bf626. Gate green: build, `go test ./internal/app`, golangci-lint 0.

PR #532, opened 2026-07-13
feat: pinned dashboard summaries (#525)

### LINKED ISSUE #525: [feat]: Summarize specific stars on the Cluster page
### Problem / motivation

Some of the quick stats on resources are how i do my 'daily check' that everything is okay, I'd like to summarize those specific ones on 'cluster' so that I don't have to go down to pods, go to jobs, go to applications, go to events, etc, just to make sure everything is okay.

### Proposed solution

Allow tagging the quick stats for each resource as 'pinned to cluster summary' (or unpinned) and show them on the Cluster summary details. That or add a new Dashboard for these stats. I don't care about _every_ stat in my cluster - for that I have the number of alerts (which I'd add to my stats list), but at a quick glance each day I wouldn't have to do my ritual travel.

### Alternatives considered

_No response_

### Area

Resource list / preview

### Additional context

_No response_

### Pre-flight checks

janosmiko:
Closes #525.

## Summary

Pin any resource type's status summary onto the cluster dashboard for a one-page daily health check - no more visiting Jobs, Applications, Pods, and every CRD list one by one.

- **Inline metric rows**: pinned summaries render as native dashboard rows directly below the Pods row - same full-width stacked bars as Nodes/Pods, worst-first status buckets, one row per dimension (Argo Applications get Health + Sync). The label column widens (cap 14) so all bars stay aligned.
- **Defaults out of the box**: when nothing is pinned, the dashboard shows Jobs, Deployments, Argo CD Applications, Flux Kustomizations, and cert-manager Certificates. CRD-backed defaults are silently skipped when not installed. `pinned_summaries: []` disables them.
- **Seeding on first pin**: the first interactive pin copies the discovery-resolved defaults into your pinned list before adding the new type, so pinning adds to what you see instead of replacing it; unpinning an active default keeps the rest.
- **Pin/unpin UX**: action menu (`x`) at the resource types level gains "Pin summary" / "Unpin summary" (chip `s`). State persists per-context or per named union set in `~/.local/state/lfk/pinned_summaries.yaml`; a `pinned_summaries` config key provides file-managed pins. Cap: 10 per scope, enforced against the merged config+state list.
- **Works for any CRD**: reuses the existing list-summary rollup, so custom resources get a phase/conditions rollup with zero per-kind code. Explicit pins for absent types show a dim "(not installed in this cluster)" row; failed lists show "(unavailable)" and log a warning.
- **Dashboard fan-out made dynamic**: the fixed 6-section accumulator became count-driven (`dashboardPartialMsg{key, total}`), with one cluster-wide list call per pinned kind, stale-accumulator eviction, and `expected = max(expected, total)` hardening against overlapping fan-outs.
- **Hardening from review**: cap checked against the effective (config+state) list; CRD status text sanitized (C0/C1/ESC/OSC stripped) at the summary choke point - covers the pre-existing preview band too; `model.DisplayNameFor` for CRD labels; eager reload after a toggle gated on the dashboard-enabled config and skipped under the union sentinel (lazy member reload picks pins up).

## Docs

README feature bullet, `docs/config-reference.md` (field row, state-files table, prose), `docs/config-example.yaml`, `docs/config.schema.json`, `docs/keybindings.md`, in-app startup tip.

## Test plan

- [x] TDD throughout: state round-trip/corrupt-file, config 4-place wiring (schema-sync + wiring-coverage tests), effective-merge order/dedupe/cap, dynamic accumulator (count-driven completion, duplicate-key dedupe, mixed-total eviction), render (pin order, zero-total row, notFound/err rows, non-mutation, label-column alignment), toggle (cap incl. config pins, save-failure rollback incl. seed, union named-set scoping, dashboard-disabled/union reload gating), default seeding (full/partial/no discovery, unpin-a-default, disabled/config-pins cases), escape-sequence sanitization (CSI, OSC-52, C1)
- [x] `go build ./...`, `go vet ./...`, `go test -race ./...` all green
- [x] `golangci-lint run` 0 issues; all files under the 800-line cap
- [x] Reviews: per-task spec+quality gates, Go review, security review (no CRITICAL/HIGH), final whole-branch review: ready to merge


janosmiko:
@coderabbitai review

PR #541, opened 2026-07-17
fix(explorer): clear quick filter when goto jumps between resource types

janosmiko:
## Summary
- A quick filter committed on one resource list stayed active after switching to a different resource list, silently hiding every row in the destination.
- All type-switch paths now share `resetFilterForTypeSwitch`: save the origin's filter for back-nav restore, then clear filter text/input, broad mode, preset state, and search highlight. Covered paths: regular descend, g-prefix goto chords (`gd`, `gv`, ...), the security-finding teleport, the per-resource findings view, and the port-forwards list. The last three previously leaked.
- Deliberately no destination-filter restore: a type switch is a fresh start; only back-nav (`navigateParent`) recalls a saved filter. Bookmark jumps and session restore keep applying their own saved filters by design.

## Test plan
- [x] `TestGotoResourceType_ClearsQuickFilter` (red before fix) + `TestGotoResourceType_DoesNotRestoreDestinationFilter`
- [x] `filter_typeswitch_test.go`: regular-nav pin + red-first leak tests for the finding teleport, findings view, and port-forwards paths
- [x] Broad-mode clear + save asserted (CodeRabbit review comments addressed)
- [x] `go test -race ./...` clean, `golangci-lint run ./...` 0 issues, `go vet ./...` clean

Tracked as Backlog TASK-839; also implements TASK-840 (`jumpToFindingResource` leak).

### DIFF
--- a/internal/app/cursor.go
+++ b/internal/app/cursor.go
@@ -410,6 +410,21 @@ func (m *Model) saveLevelFilter() {
 	m.filterMemory[key] = savedFilter{text: m.filterText, broad: m.filterBroadMode}
 }
 
+// resetFilterForTypeSwitch remembers the current list's committed filter for
+// back-nav restore, then clears all live filter/search state so the next list
+// starts clean. Every path landing on a different resource list must call it
+// BEFORE mutating m.nav — saveLevelFilter keys off the old position (TASK-839).
+func (m *Model) resetFilterForTypeSwitch() {
+	m.saveLevelFilter()
+	m.filterText = ""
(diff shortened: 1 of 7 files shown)


janosmiko:
@coderabbitai Both review comments are addressed in e1d74acc: the filter reset (now the shared `resetFilterForTypeSwitch` helper) clears `filterBroadMode`, and the goto test asserts broad mode is cleared on the destination and preserved in `filterMemory` for back-nav restore.

PR #543, opened 2026-07-17
feat(explorer): tint failed/pending rows by status (#540)

### LINKED ISSUE #540: [feat]:  whole-row background tint by pod status (failed → red, pending → yellow)
### Problem / motivation

In k9s the whole pod row is tinted by status (Failed → red background,
Pending → yellow), so broken pods pop out at a glance. In lfk only the
Status *cell* is colored; the rest of the row keeps the default
background, so a failing pod blends into the list and is easy to miss
when the Status column is narrow or scrolled out of view. Operators
scanning a long pod list today have to read the Status column row by row
instead of getting an instant visual signal.

### Proposed solution

Whole-row background tint driven by the existing status severity
classification already used for the Status cell:

- Failed (Failed, CrashLoopBackOff, Error, ImagePullBackOff, OOMKilled,
  Terminated, Degraded, …) → red row background (from `theme.error`).
- Pending / in-progress (Pending, ContainerCreating, PodInitializing,
  Terminating, Waiting, Progressing, …) → yellow row background (from
  `theme.warning`).
- Running / Succeeded / Active / Bound / Ready / unknown / empty → no
  tint (current behavior).

UX details:

- Tint applies to every column in the row, not just the Status column,
  so a broken pod stays visible even when the Status column is hidden
  via the column-visibility overlay or scrolled out of view.
- Tint is muted (background only, ~15–25% over the base background) so
  the selected-row highlight remains clearly distinguishable; a
  selected+failed row should read as "selected AND failed", not as a
  flat red block.
- Applies to all resource lists (pods, deployments, jobs, CRDs with a
  status phase, etc.) wherever `StatusStyle` is applied today.
- In `no_color` mode the tint degrades to a non-color cue: failed →
  bold row, progressing → italic row (mirroring the existing
  `StatusFailed` / `StatusWarning` no-color styles).
- Respects `appearance.min_contrast_ratio` and re-renders on theme
  switch, consistent with how other theme-dependent styles behave.
- No new config key required for the baseline behavior; a future
  follow-up could add an opt-out (e.g. `appearance.row_status_tint:
  false`) if some users find it noisy.

### Alternatives considered

_No response_

### Area

Theme / color scheme

### Additional context

_No response_

### Pre-flight checks

janosmiko:
Closes #540.

## Summary
Only the Status cell was colored, so a broken pod blended into a long list once that column was narrow or scrolled out of view. This adds whole-row emphasis driven by the existing status severity classification (the same one behind the Status cell and list summaries).

New config enum `row_status_tint` (in the `appearance` group, with a flat alias):
- **`foreground`** (default): the whole row text takes the severity color — failed = error/red, pending/progressing = warning/yellow. k9s-style, no background change.
- **`background`**: a muted severity background (~22% blended from the theme base) across the row, for a louder cue. The selection highlight stays visually distinct.
- **`off`**: previous behavior, Status cell only.

## Design notes
- Tinted non-cursor rows render from **plain cells wrapped in one row-wide style** (per-cell SGR codes would fight the tint), reusing the cursor row's established plain-cells + `RenderOverPrestyled` mechanics. The `plainRestartsCell` preprocessing was extracted and shared.
- Cursor/selection styling always wins — the cursor row never tints.
- `no_color` mode degrades to bold (failed) / italic (progressing).
- Styles re-derive on theme switch (defined in all three theme files); the row cache is keyed on `ThemeRev`. `row_status_tint` itself is startup-only (documented at the fingerprint struct).

## Test plan
- [x] `row_tint_test.go`: `RowTintForStatus` across all modes + `no_color`, `blendHexToward` math, and renderer SGR assertions (foreground/background/off), red-first
- [x] Config wiring: `config_wiring_test.go` (default + coverage map) and `config_appearance_test.go` (group-wins, invalid-fallback) — the four schema/wiring forcing-function tests pass
- [x] `go test -race ./...` clean, `golangci-lint run ./...` 0 issues, `go vet ./...` clean, ui coverage 87.3%

## Docs
Schema (top-level + appearance), config-reference table, config-example, README.

Reviewed by `go-reviewer` (approved, no CRITICAL/HIGH). Tracked as Backlog TASK-841.


janosmiko:
Updated per review feedback (commit d0563481):

- **Selection fix**: selecting a failed/pending row no longer erases its status signal — the severity color rides on the row foreground over the selection highlight (no-color falls back to underline for failed, since selection is already bold).
- **Progressing color fix**: the tint now matches the existing Status cell color (primary/blue), not the warning color — locked by an invariant test so the row tint and the cell can never disagree. (Thanks for catching this @janosmiko.)
- **Foreground is now a fallback, not always-on**: no tint while the Status column is visible; color only the Name cell when Status is hidden; whole-row only when Name is also hidden. Background mode still tints the whole row.
- **CodeRabbit**: the invalid-value warning now logs the accepted options, not the raw input.

Docs/schema updated to match.

janosmiko:
Selection behavior fixed per your feedback (commit 897aa094):

- **Background mode**: a selected failed/pending row now **keeps its status background** instead of swapping in the blue selection background. The cursor is marked with the existing green checkmark + bold text (reusing `SelectionMarkerStyle`), so the 'failed' background never disappears on selection.
- Same-class bug fixed: a **multi-selected** (non-cursor) tinted row was losing its background right after the checkmark when there's no cluster-color tile — the tint is now re-asserted after the marker.
- Foreground/off modes unchanged.

Row-tint helpers moved to `row_tint.go` (styles.go hit the 800-line cap). Full `-race` suite / lint / vet green.

janosmiko:
Cursor visibility on tinted rows fixed (commit de3a4f8e):

- The cursor row on a status-tinted row now uses a **blend of the status background and the selection color** (`rowTintCursorBlend`), so a selected failed row reads as both 'failed' (status hue) and 'cursor' (selection hue) instead of the cursor highlight disappearing. Progressing rows blend toward a brighter blue.
- No-color mode falls back to the reverse-video selection (with underline) so the cursor stays distinct there too.

New style globals `RowTint{Failed,Progressing}CursorBg` in all three theme files; re-derived on theme switch. Coverage added for the no-color/background path. Full `-race` suite / lint / vet green.

janosmiko:
@coderabbitai review

janosmiko:
@coderabbitai Both inline comments addressed:
- Log redaction (config_row_tint.go): fixed in d056348 — the invalid-value warning logs the accepted options, not the raw input.
- Docs contract (schema + config-reference): fixed in 6a62fc4b — documented the foreground whole-row fallback when both Status and Name are hidden, the cursor blend in background mode, and the underline cue for a selected failed row in no_color mode.

PR #570, opened 2026-07-31
feat(delete): selectable cascade policy for delete and force delete

janosmiko:
## Why

`DeleteResource` sent an empty `metav1.DeleteOptions{}`, so the API server applied its own per-resource default. For `jobs.v1.batch` and `replicationcontrollers.v1` that default is still `OrphanDependents` (upstream `pkg/registry/batch/job/strategy.go`), so **deleting a Job from lfk left its pods running**. `kubectl` avoids this by always sending `propagationPolicy` explicitly.

Force delete had the same gap from the other direction: it shelled out to `kubectl delete` with no `--cascade`, so it always ran kubectl's implicit `background` no matter what was configured.

Upstream asserts the exhaustive legacy-orphan set in `test/integration/apiserver/apidefinitions/default_behaviors_test.go`:

```go
orphanDependentsExempt := sets.New("jobs.v1.batch", "replicationcontrollers.v1")
```

## What

Every delete now sends an explicit policy, and the policy is selectable per delete with `Tab`.

```
┌ Confirm Delete ────────────────────┐
│  Delete job/backup-nightly?        │
│  Cascade: Orphan (dependents kept) │   ← warning-styled
└────────────────────────────────────┘
 Enter/y confirm  tab cascade  Esc/n cancel
```

| Policy | Effect |
|---|---|
| `Background` | Delete now; GC removes dependents (kubectl's default, and ours) |
| `Foreground` | Keep the object until every dependent is gone |
| `Orphan` | Leave dependents running |
| `None` | Send no policy; the API server applies its per-resource default |

- Applies to single **and** bulk delete, and to the force-delete typed confirmation.
- `Orphan` and `None` are warning-styled and annotated, since both can leave workloads running.
- `delete_propagation_policy` sets the starting value (default `background`). An invalid value falls back to `background` rather than being left unset, so a typo cannot reopen the orphan trap.

### Deliberate asymmetry: force delete omits `None`

`kubectl` cannot express "send no policy" — omitting `--cascade` makes it send `background` anyway, and `--cascade=none` is rejected outright. So the force cycle is Background/Foreground/Orphan and a configured `none` clamps to `background`. Offering a `None` that silently behaved as `Background` seemed worse than not offering it.

`Tab` is inert where it would do nothing: Force Finalize, Finalizer Remove, Disrupt, helm uninstall, and Longhorn nodes (which delete via a webhook-aware API call, not kubectl).

### How this compares to other tools

| Tool | Default | Selectable? |
|---|---|---|
| **k9s** | Background | Yes — Background/Foreground/Orphan/None |
| **kubectl** | Background | Yes, `--cascade=` |
| **ArgoCD** | Foreground | Yes, `PrunePropagationPolicy` |
| **Lens, Headlamp** | Unset; Background hardcoded for Jobs only | No |
| **Rancher Dashboard** | Unset, no propagation param | No |

Headlamp and Lens both special-case Jobs to `Background`, which is independent confirmation that the server-side Job default is a recognized footgun.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race ./...` — exit 0, 23 packages
- [x] `gofmt -l internal/` — clean
- [x] Fake dynamic client asserts `DeleteOptions.PropagationPolicy` per policy, and **nil only for `None`**
- [x] Every policy — including unset and garbage — yields a `--cascade` value kubectl accepts
- [x] TUI: Tab cycling, reseed-on-open, non-delete/helm/Longhorn gating, narrow-width rendering at 30/40/50/120 cols
- [x] Config: schema/wiring/coverage forcing-function tests
- [ ] Manual: delete a Job in a real cluster with each policy and confirm pod behavior

Reviewed by `go-reviewer` and `security-reviewer`; no CRITICAL/HIGH. Findings fixed: an unused `OrphansDependents()` with `Orphan` visually interchangeable with safe policies; a `deletePropagationOrder[:3]` slice that shared a backing array (an `append` would have silently overwritten the `None` entry); a missing cascade field in the bulk force-delete log.

## Notes for the reviewer

- Docs are complete as of the second commit: `docs/keybindings.md` (both dialogs + policy tables), `docs/config-reference.md`, `docs/config-example.yaml`, `docs/config.schema.json`, `docs/views-and-overlays.md`, the README's Resource Operations list, and the in-app help. An earlier revision of this description said the keybinding docs were deferred; that is no longer true.
- Three files sat at exactly the 800-line revive cap and had to be relieved to make room: `applyWatchIntervalConfig` moved to its own file, the confirm-overlay renderers were extracted from `view_overlays.go`, and the duplicated conditional-hint logic became `cascadeConfirmHints`. `app.go` is still at exactly 800 — the next field added to `Model` will trip lint.
- Force delete with `Orphan` is the widest-reaching combination in lfk: bulk force delete strips finalizers first, so the owner vanishes while its pods keep running with no owner reference. The security review flagged this as a design judgment; the decision was to keep the typed `DELETE` as the gate rather than add a second acknowledgement for one policy value, and document the risk instead. Worth a second opinion.


janosmiko:
Addressed all four review comments in 5ebcef52.

**`internal/ui/config_delete_policy.go` (Major)** — fixed, with a caveat on the diagnosis. `LoadConfig` is called exactly once (`main.go:112`, no reload path), so the global was always still at its compiled `background` default when this ran and the mismatch could not occur in production. The order-dependence was real though, and reachable in tests since `TestLoadConfig_AllSettingsWired` sets `orphan` in the same process — the existing tests only passed because they snapshot and restore the global. Both fallbacks now assign Background explicitly, and `TestDeletePropagationPolicy_ResetsStaleValue` loads `orphan` first, then asserts invalid / omitted / empty all reset. Those subtests fail without the fix.

**`internal/ui/config_load.go` (Minor)** — fixed. Correct catch; the comment was written before `none` existed and never updated. Now lists all four and notes the force-delete clamp.

**`README.md` (Minor)** — fixed. Also a correct catch, and the bullet contradicted itself: it documented `None` as sending no policy and then claimed every delete sends an explicit one. Reworded to say Background is the explicit default and `None` deliberately opts back into the server-side default.

**`internal/ui/overlay_confirm_test.go` (Nitpick)** — taken. Two independent configs instead of mutating one.

On the **Docstring Coverage** pre-merge warning (66.67% vs 80%): not addressing it. The only undocumented symbol in the diff is `cycleIn`, a three-line private helper whose name states what it does. This repo's documentation guidelines (`~/.claude/rules/common/documentation.md`, mirrored in CLAUDE.md) specify "Default: none — add only when WHY is non-obvious" and flag comments that restate the function name as an issue to remove. Adding docstrings to hit the threshold would violate the project's own documented style, so the threshold is the wrong target here.

PR #578, opened 2026-08-05
feat: context-aware which-key panel, and a reworked help screen

janosmiko:
Adds a which-key panel that lists the hotkeys actionable *right now*, in the explorer and eight fullscreen viewers, and reworks the help screen to match.

Closes TASK-848, TASK-849, TASK-851.

## What this adds

**The panel.** `?` opens a bottom-anchored panel of the keys that apply to the current row and view. Entries carry availability predicates read off their real handlers, so a key that would no-op is not advertised. Layout is a port of neovim's which-key: uniform full-width columns, column-major fill, grow-to-fit with scrolling, key glyphs (`⌃ ⇧ ⌥`, `⇥ ␣ ⌫`, arrows) gated on the existing `icons` mode.

**Grouping.** Entries cluster by category with descriptions coloured per group and a legend on the bottom line. Pressing `?` again toggles to pure key ordering; the mode persists for the session and its default is `which_key_grouped`.

**Coverage.** Explorer plus YAML, log, describe, diff, API Explorer, Object Explorer, Log Top and the fullscreen event viewer. Exec mode deliberately gets no catalog — it forwards every keystroke to the PTY, so a panel there would steal `?` from the shell.

**Help screen.** One binding per line, right-aligned symbol keys, dimmed `/` between alternatives, 39 prose-only rows removed, and search/filter that still match textual key forms against glyph-rendered rows.

## Breaking changes

- **`?` no longer opens the help screen** in the explorer or any viewer with a panel. `f1` opens help everywhere.
- **`restart` and `exec` keybindings removed.** They were dead — declared, defaulted and documented, but no dispatcher compared against them. Unknown config keys are silently ignored, so an existing `restart:` line keeps working exactly as before, which is to say it does nothing.

## Config

| Key | Default | Effect |
|---|---|---|
| `which_key_enabled` | `true` | Gates the panel and the `g` popup |
| `which_key_leader` | `?` | Arms the panel |
| `which_key_grouped` | `true` | Start grouped or in key order |
| `which_key_leader_delay_ms` | `0` | Reveal delay |

## How this was verified

Reviewed in six independent passes across the branch — coherence, geometry, test integrity, predicate accuracy (twice), and integration. Predicate audits traced every catalog entry to its handler's dispatch site; the eight viewer catalogs produced one defect between them.

The suite is built around forcing functions rather than spot checks, each proven to fire by injecting the defect it catches:

- no two simultaneously-available entries may resolve to the same key
- every entry appears in the rendered output at some scroll offset
- every `ui.Keybindings` field is registered or explicitly excluded with a reason
- the help and panel taxonomies must agree, or CI fails
- every literal key in a help row must be a real `case` operand in that view's dispatcher
- allocation ceilings on the availability pass and the render path
- geometry invariants swept across every width, height, icon mode and scroll offset

Along the way this found and fixed a pre-existing bug: `ctrl+space` range-selection had been dead since the Bubble Tea v2 migration, because `SelectRange` defaulted to `ctrl+@` while v2 decodes the real keypress as `ctrl+space`.

## Known limitations

- Four overlay views (can-I, error log, sync-wave, network policy) are `overlayKind` rather than `viewMode` and return before the leader dispatch, so they have no panel.
- About two dozen viewer-local actions match hardcoded keys and ignore their config field; rebinding those moves the key in the explorer only. Both surfaces report the split honestly. Tracked in TASK-852.
- Rebinding `jump_top` disables the built-in `goto` chords, whose defaults hardcode the `g` prefix. The popup no longer advertises the unreachable ones.
- Nerd Font glyphs are East-Asian-Ambiguous; a terminal set to `ambiguous=wide` will misalign columns. `icons: simple` is the escape hatch.

## Test plan

- [ ] `?` in the explorer: panel appears, entries grouped and coloured, legend centred
- [ ] `?` again toggles to key ordering; `esc` closes; reopening remembers the mode
- [ ] `space` still multi-selects; `space j space j` selects two rows
- [ ] `ctrl+space` range-selects from the anchor
- [ ] `?` in YAML, logs, describe, diff, both explorers, Log Top, event timeline
- [ ] `?` in exec reaches the shell
- [ ] `f1` opens help; `/` searches; `f` filters
- [ ] `--read-only` hides destructive entries
- [ ] 80x24 and a narrow tmux split render without overflow


janosmiko:
Thanks — nine of the ten applied in `d781057c..837e39fb`. One I'm declining, with reasoning below.

**Applied**

- **Diff context memo** (`whichkey_diff.go`) — content-keyed cache on `left`/`right`/`foldState` rather than a generation counter, since the fold keys mutate `foldState` in place and a counter is one missed bump away from a stale line list. Benchmarked at 400 lines/side: **673µs → 1.27µs, 1.49MB → 1.86KB, 429 → 3 allocs/op**. Validated by deliberately aliasing the cache and confirming two tests catch it.
- **`q` labels** — checked all ten viewers against their handlers. Object Explorer and API Explorer were wrong (`exitObjectExplorer` / `handleExplainKeyQ` close unconditionally at any depth); the other eight are correct. The rule applied: "Back" is only wrong where the view has a level stack, because there `h`/`backspace`/`esc` genuinely step back and `q` does not. Reading the help rows for this also turned up a real error next door — the Object Explorer section merged `q/esc` into one row.
- **Width measurement** — the sweep found two more instances beyond the one reported, both in `whichkey_legend_test.go` and both worse: `len(body)` in **bytes** against a column budget. Measured on a rendered 120x40 panel, rune-vs-cell delta is 0 but byte-vs-cell delta is **224**, so that width guard was ~224 columns too permissive and effectively dead.
- **All five documentation findings**, each verified against source rather than taken from the review.

**One deviation:** I wrote `help: "f1"`, not `"F1"`. `NormalizeKeybinding` leaves non-chord strings untouched and only upper-cases behind `shift+`; keys are matched against `msg.String()`, which is `"f1"`. `docs/keybindings.md` states case is significant, so `"F1"` would silently never fire.

**Declining: `whichkey_leaderkey_test.go:92.`** The finding says the test sets `ConfigWhichKeyLeaderDelayMs` while `armWhichKey` reads `ConfigWhichKeyDelayMs`. `TestViewers_LeaderArmsThePanelAndF1OpensHelp` exercises viewer modes through `handleKey` → `handleViewerWhichKeyLeader` → `armWhichKeyLeader` (`whichkey_leader.go:95`), which reads exactly the field the test sets. `armWhichKey` (`whichkey.go:234`) is the `g`-prefix goto popup and is never on that path. The two similarly-named arming functions read different config fields by design.

janosmiko:
Both findings from the full re-review were real and are fixed in `1c187413`, `9b597b02`.

**`GotoChordReachable` accepting unreachable chords** — a genuine bug in code added earlier in this branch to prevent exactly this class. `handleGotoChord` builds `JumpTop + key` from one keypress, so `gAA` can never be produced, but the old check only tested the prefix and that *something* followed.

The fix measures the tail in keypresses rather than bytes, via a new `ui.IsSingleKeypress` living next to `NormalizeKeybinding`. "One keypress" is not one rune — `gctrl+p`, `gtab`, `gf12`, `gctrl++` are all legitimate and stay reachable, so a naive length check would have broken every modifier chord. Cases added on both sides of the line: rejects `gAA`, `gpp`, `gctrl+pp`, `ga+b`; accepts `gctrl+p`, `gctrl+alt+y`, `gctrl++`, `g+`, `gtab`, `gpgup`, `gf12`, `gspace`.

This also removed the second copy of the rule you'd implicitly pointed at — `applyGotoTargets` had its own `len([]rune(chord)) != 2` check, now calling the shared predicate. Side effect, deliberate: user `goto_targets` now accept `gctrl+p`/`gtab`, which the dispatcher could always build but the two-rune rule rejected.

**The unused `byKey` map** — correct, the test would have passed with every cell assigned the same wrong group. Now compares each cell against the catalog's own group for that key.

One correction worth recording: proving it required injecting into `whichKeyLeaderCells`' copy path, not the catalog. A wrong-but-declared group in the catalog is invisible to any panel-vs-catalog comparison, since both sides read the same source — I ran that injection and it passes. Injecting a per-entry swap in the copy path is the discriminating case: the old test passes, the new one fails naming seven entries. A blanket swap doesn't discriminate either, because the old test's `len(groups) < 2` precondition already caught that.

The duplicate-key concern doesn't apply here: `wkDropAmbiguousKeys` drops both entries when two survivors share a key, so `availableWhichKeyActions()` is unique by key by construction. The test now fails loudly if that invariant ever breaks rather than silently overwriting.

janosmiko:
Correct, and fixed in `bfae4800`. Verifying it turned up a second instance of the same bug in the neighbouring parser.

Only `ctrl++` spells a modified `+`: it splits to `[ctrl "" ""]`, so the component *before* the empty key is empty too. Both chord parsers dropped that component unconditionally, so `ctrl+`, `ga+` and `ctrl+alt+` all resolved to the `+` key.

**`IsSingleKeypress`** — as reported. `GotoChordReachable("gctrl+", "g")` returned true, so the popup could advertise a chord the dispatcher can never build. Regression cases added for `gctrl+`, `gctrl+alt+`, `gga+`; verified they fail against the old code.

**`NormalizeKeybinding`** — not reported, found by probing the same input through the neighbouring parser, and worse. `NormalizeKeybinding("ctrl+alt+")` returned `"ctrl++"`: a malformed binding silently rewritten into a *different, working* one. That directly contradicts the function's own doc comment, which promises malformed input is "returned untouched rather than silently reordered into something that also would not match". A user with a trailing-plus typo would have got the literal `+` key bound under ctrl, with no warning.

Both now leave malformed input untouched. Confirmed unchanged for the valid forms: `ctrl++` → `ctrl++`, `+` → `+`, `shift+x` → `X`, `ctrl+p` → `ctrl+p`.

janosmiko:
Correct on both counts, and the observation about `ga+` is the important half. Fixed in `8acf4b4f`.

`NormalizeKeybinding("alt+ctrl+")` returned `"ctrl+alt+"` — the previous guard stopped the `+`-key misreading, but a malformed binding whose modifiers are all *valid* still fell through to the reordering pass. Now it bails as soon as the key is empty, before reordering.

The part worth recording: **the three cases I added with that guard all passed for reasons other than the rule.** `ctrl+` has one modifier so needs no reordering; `ctrl+alt+` is already canonical; `ga+` early-returns on the unknown modifier, exactly as you noted. None of them reached the reordering path, so none could have caught this. `alt+ctrl+` — valid modifiers, non-canonical order — is the only shape that does.

That is a vacuous pass, and it was in a test written specifically to close the previous round's finding. Verified the new case fails against the pre-fix code before committing.

PR #579, opened 2026-08-06
fix: scope Ctrl+C bulk-action cancel to the tab that started it

janosmiko:
## Problem

Start a long-running bulk action (e.g. select all ArgoCD Applications, `x` -> `s`), switch to a different tab, press `Ctrl+C`: the tab does not close, the action is "cancelled" instead.

The scheduler registry was global. `HasActiveMutations()` returned true for any in-flight mutation anywhere, so the `Ctrl+C`/`Esc` gate in `handleExplorerNavKey` swallowed the key on every tab.

## Fix

- `TabState.uid` — stable per-tab identity (tab index shifts on close/reorder, so it cannot serve).
- `scheduler.Task.Owner` — every task records the tab that started it. Tagged at all three mutation entry points: `trackBgTask`, `scheduleK8sCall`, `StartCancellable`. Owner `0` means unowned and matches every tab.
- `Ctrl+C`/`Esc` uses `HasActiveMutationsOwnedBy` / `CancelMutationsOwnedBy`. Foreign work no longer intercepts the key.
- Closing a tab calls `DisownTasks(uid)` so its still-running work stays cancellable from the remaining tabs — UIDs are never reused, so the task would otherwise be permanently orphaned.

## Test plan

- [x] `go test -race ./...`
- [x] `golangci-lint run ./...` (0 issues)
- [x] New tests: `internal/app/tab_scoped_cancel_test.go` (foreign tab closes, owning tab cancels, closing owner disowns, UID uniqueness) and owner-scoping unit tests in `internal/app/scheduler/registry_test.go`
- [x] Verified the new tests fail when the ownership check is removed
- [ ] Manual: bulk ArgoCD sync on tab 1, switch to tab 2, `Ctrl+C` closes tab 2 and the sync keeps running

## Notes

Out of scope: `bulkSyncArgoApps` / `bulkRefreshArgoApps` register via `trackBgTask` with no cancel func, so even on the owning tab `Ctrl+C` shows "Cancelling..." without stopping them. Separate fix.

### DIFF
--- a/docs/keybindings.md
+++ b/docs/keybindings.md
@@ -985,7 +985,7 @@ Invalid config values are dropped at startup with a warning in the error log.
 | `[` | Previous tab |
 | `}` | Move current tab right (shift+]) |
 | `{` | Move current tab left (shift+[) |
-| `Ctrl+C` | Close current tab (quit if last tab) |
+| `Ctrl+C` | Close current tab (quit if last tab). Cancels a bulk action first, but only one started on this tab |
 
 ## Read-Only Mode
 
--- a/internal/app/app_init.go
+++ b/internal/app/app_init.go
@@ -137,6 +137,7 @@ func NewModel(client *k8s.Client, opts StartupOptions) Model {
 		reqCancel:                  reqCancel,
 		middleTableRenderer:        ui.NewTableRenderer(),
 		tabs: []TabState{{
+			uid:                nextTabUID(),
 			nav:                model.NavigationState{Level: model.LevelClusters},
 			namespace:          defaultNS,
 			splitPreview:       ui.ConfigSplitPreview,
--- a/internal/app/app_types.go
+++ b/internal/app/app_types.go
@@ -443,6 +443,11 @@ type hpaScaleState struct {
 
 // TabState holds per-tab navigation state so each tab is fully independent.
 type TabState struct {
+	// uid identifies this tab for the lifetime of the process. Background
+	// work records it so Ctrl+C only cancels operations started from the
+	// tab the user is on (tab index shifts when tabs close or reorder).
+	uid uint64
+
 	// needsLoad is true for tabs restored from a session file that have not
 	// yet had their items loaded.  When loadTab detects this flag it triggers
 	// a full refreshCurrentLevel instead of the lighter loadPreview.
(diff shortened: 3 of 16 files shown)


janosmiko:
Addressed in 5299e2d8.

**Fixed (both Major):**

- **Dedupe key across tabs** — `Owner` is now part of the `(Kind, Name, Target)` signature in `startWithPriority`. Two tabs running the same operation are two operations; collapsing them left the loser running without a cancel func. Test: `TestSameSignatureTasksFromTwoTabsBothSurvive`.
- **Queued task keeps a dead owner** — real gap: `Submit` carries `req.Owner` and the visibility task registers later in the worker, after the tab may be gone. Rather than threading ownership through the queue, the Registry records closed-tab UIDs and strips the owner at `startWithPriority` — one choke point that covers queued, picked, and any future path. The set is bounded (`disownedCap = 64`); queues drain in seconds and tabs cap at 9, so an evicted UID can no longer have pending work. Tests: `TestStartStripsOwnerOfAlreadyClosedTab`, `TestDisownedSetIsBounded`.

Both fixes were verified to fail their tests when reverted.

**Not taking (nitpicks):**

- **Lazy assignment in `currentTabUID`** — deliberate. It is a pointer-receiver lazy init, idiomatic in Go, and it is the safety net: a tab-creation path that forgets a UID gets a real one instead of silently falling back to owner `0`, which would quietly restore the global-cancel bug. Making it read-only moves that failure mode from "impossible" to "silent".
- **Copy-on-write in `DisownTasks`** — `r.tasks` is `map[uint64]*Task` and every mutator in the file (`UpdateProgress`, `Finish`) writes through the pointer under `r.mu`. `Snapshot` returns values, so no reader holds the pointer. Copying here would be inconsistent with the surrounding code for no safety gain.
- **`basePush80Model()` + `stripANSI(view)`** — that guideline covers rendering assertions. These are state/dispatch tests (which tab closed, which cancel func fired); `threeTabModel` is the established fixture for tab-close tests in `tab_close_test.go`. Rendering is not what is under test.

janosmiko:
Fixed in 58fcc451.

You're right that the cap was unsound: there is no point at which delayed registration becomes impossible — a queued task can sit behind arbitrarily long work — so any count- or time-based eviction can drop a UID that still has something pending.

The cap is gone; `disowned` is kept for the process lifetime. Growth is one 8-byte key per tab the user closes, so even a pathological session (closing a tab every 10s for 8 hours) lands around 100 KB. That is well below the level worth an eviction policy, and the policy was the bug.

Regression test added as requested: `TestStartStripsOwnerAfterManyLaterClosures` disowns owner 1, closes 199 more tabs, then registers owner 1's queued work and verifies another tab can still cancel it. Verified it fails when the disowned-set record is removed.

PR #588, opened 2026-08-09
feat: show the cluster schema description of the field under the cursor

janosmiko:
Closes TASK-863.

`Ctrl+K` opens a side pane in the YAML viewer and the Object Explorer showing what the connected cluster says about the field under the cursor. The pane follows the cursor as it moves.

The text comes from `kubectl explain` against the current context, not from a bundled schema, so it matches the cluster version and CRD fields resolve the same way built-in kinds do.

```
┌─ pod.yaml ──────────────────┐  SCHEMA <string> spec.dnsPolicy
│ spec:                       │ ╭─────────────────────────────╮
│   dnsPolicy: ClusterFirst   │ │ Set DNS policy for the pod. │
│   restartPolicy: Always     │ │ Defaults to "ClusterFirst". │
│   containers:               │ │                             │
└─────────────────────────────┘ ╰─────────────────────────────╯
 j/k scroll  ...  ctrl+k schema  q/esc back
```

## Design notes

**Cost.** One `kubectl explain` is a process spawn and a round trip. The pane waits 250ms for the cursor to rest before asking, so holding `j` spawns nothing. Every answer is cached per context, api-version, resource and field path, bounded at 256 entries with FIFO eviction. An empty description is cached too: a field the schema leaves undocumented is an answer, and re-asking on each visit would spawn `kubectl` for nothing.

**Stale replies.** Each fetch carries a request number. Moving the cursor bumps it, so a reply for a field the user has left is dropped instead of painting the wrong text under the new one.

**Layout.** The pane is built on the same shape as the log viewer's structured preview (`RenderLogPreviewPane`) and joined horizontally, with one full-width hint bar under both panes. It takes columns, not rows. `splitFieldDocWidth` uses the same bounds as `splitLogPreviewWidth`, so the two side panes come out the same width on the same terminal. Below 90 columns `Ctrl+K` reports that the terminal is too narrow rather than toggling a pane nobody can see.

**`Ctrl+K`, not `K`.** `K` is `kb.PreviewUp`, and the Object Explorer scrolls its preview with `J`/`K`. No existing binding moved. Configurable via `keybindings.field_doc`.

**Errors.** `kubectl explain` prints the `KIND`/`VERSION` preamble even when it fails, so the pane showed `exit status 1: KIND: Pod ...` wrapped around the one useful line. `parseExplainError` keeps the `error:` line alone.

## Note on AC #3

The task asks for the "cluster OpenAPI schema". The text comes from the live cluster through the `kubectl explain` path — discovery never fetched an OpenAPI schema; `openapi` appears nowhere in the code, only in `go.sum`. The intent is met (the cluster's own descriptions, not a bundled copy), but the criterion's wording describes a mechanism that does not exist.

## Docs

README features and hotkey table, `docs/keybindings.md` (both viewer sections and the config block), `docs/config-reference.md`, `docs/config-example.yaml`, `docs/config.schema.json`, `docs/views-and-overlays.md`, in-app help, and the which-key catalogs for both viewers.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race ./...` — exit 0
- [x] 60 new tests: parser, error extraction, cache bounds and eviction, key wiring, `K` still scrolls the OE preview, pane rendering in every state, width split, hint-bar parity, joined view keeps the exact terminal box
- [ ] **Not verified:** never run against a live cluster. The `kubectl explain` subprocess itself is not exercised by any test — only its output parsing. Worth a manual pass on a Pod and on a CRD before merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ Cloudsmith is the only fully hosted, cloud-native, universal package management 
 - **Monitoring dashboard** with active Prometheus/Alertmanager alerts (`@` key), configurable endpoints per cluster
 - **API Explorer** for interactively browsing resource structure (`I` key) with recursive field browser and an ASCII-art tree view (`T`, sticky across navigation; `Space` folds branches)
 - **Object Explorer** for drilling into the selected resource's live object (`O` key); arrays expand into indexed elements, so recursive status trees (e.g. `.status.steps[].steps[]`) are walkable; live-refreshes under watch mode (`w` to pause, `R` to refresh manually); `T` toggles a tree view that expands the whole subtree with ASCII-art guides, `Space` folds branches
+- **Schema footnotes** (`Ctrl+K`) show the cluster's own description of the field under the cursor in a pane at the bottom of the YAML viewer and Object Explorer; the pane follows the cursor, reads the connected cluster so CRDs resolve like built-in kinds, and caches what it reads
 - **Namespace selector** overlay with type-to-filter
 - **All-namespaces mode** (enabled by default)
 - **Local cluster management** — create, list, and delete `kind` clusters; create, list, start, stop, and delete `k3d` and `minikube` clusters (kind has no native start/stop) from inside lfk via the `Ctrl+N` manager overlay.
@@ -282,6 +283,7 @@ Namespaces are **not** a navigation level. The current namespace is shown in the
 | `Ctrl+G` | Finalizer search and remove |
 | `I` | API Explorer (browse resource structure interactively) |
 | `O` | Object Explorer (browse the selected resource's live object as a drill-in tree) |
+| `Ctrl+K` | Schema footnote for the field under the cursor (YAML viewer, Object Explorer) |
 | `U` | RBAC permissions browser (can-i) |
 | `T` | Open theme selector |
 | `:` | Command bar: resource jumps (`:pod`, `:dep`), built-ins (`:ns`, `:ctx`, `:set`, `:sort`, `:export`), kubectl (`:k get pod`), shell (`:! cmd`) |
@@ -403,6 +405,7 @@ All search and filter inputs support three modes, auto-detected from the query s
 - Walk any resource's live object with `O` (Object Explorer): `r` finds keys recursively, `T` expands an ASCII tree, `y` copies the field path
 - Forget `kubectl explain` - `I` opens the API Explorer, and `n`/`N` searches auto-drill into nested fields
 - In the YAML viewer, press `O` on a line to jump into the Object Explorer at that attribute, or `I` to see its schema
+- `Ctrl+K` opens a footnote with the schema description of the field under the cursor, and keeps it in step as you move
 - Fold YAML sections with `z` (`Z` folds all); edit the resource in your `$EDITOR` with `Ctrl+E`
 - Every viewer speaks vim: counts (`100j`, `42G`, `5n`), visual selections (`v` / `V` / `Ctrl+V`), and text objects (`viw`) work everywhere
 - Make noisy logs readable with `P` - the structured preview parses JSON, logfmt, klog, zap, nginx, envoy, Java, and postgres lines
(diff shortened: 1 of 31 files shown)


janosmiko:
All four review points addressed in 1f87b144. Each was checked against the code first; all four were real.

**Bound and cancel the subprocess** — fixed. The call now runs under the request context with a 15 second deadline (`fieldDocFetchTimeout`), so an unreachable cluster or a credential plugin waiting on input cannot hold a scheduler worker. You were right that retargeting only dropped the reply.

**Raw output in the log** — fixed. Only the target, api-version, and exit status are logged. On an auth failure `CombinedOutput` carries whatever the credential plugin printed, which is the case that matters. The parsed message still reaches the pane.

**Tautological path-tail test** — fixed, and it was hiding a real bug. The description contained "port", so the assertion passed while the title dropped the path entirely at narrow widths: `TitleStyle.MaxWidth` cuts from the right, which removes exactly the leaf. The title comment claimed it cut from the front; it never did. It does now, via `TruncateStart`, and the test asserts on the title row with a description that omits the word.

**Path as `[]string`** — fixed for display. The title is now built from the original segments through `formatObjectPath`, so an array index reads as `[0]` and a segment holding a dot stays one segment. The dot-joined form is kept for the kubectl target and the cache key, which is the only thing it was ever for; it is never split back apart.

Two notes for the record:

- The same missing-context and raw-output patterns exist in the pre-existing `execKubectlExplain` (`internal/app/commands_exec_pod.go:432`). Left alone here to keep this diff to its own scope; worth a follow-up.
- Still not verified against a live cluster. No test runs the `kubectl explain` subprocess, only its output parsing.

Gate: build, `go vet`, `golangci-lint` 0 issues, `go test -race ./...` exit 0.

janosmiko:
Fixed in fd4c5487. The finding was correct: the deadline bounded one call, but retargeting only marked the reply stale, so the process ran on and held a scheduler worker. The pool is small enough for that to matter on a slow cluster.

`fieldDocState` now holds the request cancel func. It fires on four paths:

- a new target replaces the request (`showFieldDoc`)
- a cache hit answers without a fetch, making anything still running stale
- the reply lands, so the context is released rather than left to the parent
- the pane closes (`reset`)

The request context is the parent of the 15 second timeout context, so either path ends the process.

Five tests cover it: retarget, close, completion, cache hit, and that cancelling with nothing running is a no-op.

Gate: build, `go vet`, `golangci-lint` 0 issues, `go test -race ./...` exit 0.

janosmiko:
Integration half done in d643631a. The E2E half I am declining, with a reason.

**Integration — added.** A stub `kubectl` goes first on PATH, writes its argv to a file, and prints whatever the case needs. Seven tests now cover the subprocess path that nothing touched before:

- argument construction: target joins resource and field path, `--context` present, `--api-version` present for a CRD and **absent** for a core resource
- the parsed description and field type, and that the reply carries the request number it answers
- the trimmed error output, end to end through a stub that exits 1 after printing the `KIND`/`VERSION` preamble
- a missing `kubectl`
- cancellation of a running process

The cancellation test lets the process start before cancelling, so it proves the running process dies rather than proving an already-cancelled context never spawns one. I checked it against the bug: with the timeout context parented to `Background` instead of the request context, it fails after 30 seconds. It passes in 0.34s with the fix.

**E2E — not done.** The repo has no E2E harness: no `e2e` build tags, no test directory, no kind or k3d in CI. Cluster-backed tests are a new capability for this repo, not part of this feature, and bolting a first one onto this PR would land infrastructure nobody reviewed. Worth its own task if you want it.

So the live-cluster caveat on this PR stands, but it is now narrower: the subprocess contract is covered, and what remains unverified is whether a real cluster's `kubectl explain` output matches the shapes the stub reproduces.

Gate: build, `go vet`, `golangci-lint` 0 issues, `go test -race ./...` exit 0.

janosmiko:
Both points fixed. 12ab4afb and 26af7532.

**The CI failure was a real bug, and the review comment led straight to it.** Killing kubectl does not kill what kubectl started, and `CombinedOutput` reads to EOF, so a surviving grandchild keeps the output pipe open and the worker stays blocked on a process that is already dead. A credential plugin is the real-world case. CI named it in its own cleanup line: `Terminate orphan process: (sleep)`. Fixed with `cmd.WaitDelay`, which gives the pipes two seconds after the kill and then gives up on them. `build-and-test` went from fail at 2m56s to pass at 2m27s.

**The cancellation assertion** was weak exactly as described: 30 seconds would also have passed if cancellation were ignored and the 15 second deadline ended the fetch. It now waits for the stub to write its argv, so the process is really running, measures from the `cancel()` call, and allows less than half the deadline.

One part of the suggestion I did not take: `exec sleep 60`. That removes the grandchild, which is precisely the shape that broke CI and the reason `WaitDelay` exists. Keeping the backgrounded child is what makes the test cover it. With the child kept, the test now exercises that path on macOS as well.

Verified both ways rather than assumed: 2.4s with `WaitDelay`, failing at 7.9s without it.

Gate: build, `go vet`, `golangci-lint` 0 issues, `go test -race ./...` exit 0.

PR #589, opened 2026-08-09
feat: show the blast radius in destructive confirm dialogs

janosmiko:
Closes TASK-856.

The delete, drain, scale-down, and bulk-delete dialogs now state what the action costs before you answer: how many pods go, how many ready replicas are left, and which PodDisruptionBudgets it spends. A budget the action would breach gets its own warning-styled line naming the shortfall.

```
Confirm Delete
Delete web?
Cascade: Background
Blast radius: 2 pods, 3 of 5 ready after, PDB web-pdb 2 -> 0 allowed
```

Cordon is deliberately out of scope. It evicts nothing, so its budget headroom does not change and a line reading `2 -> 2` would teach people to ignore the feature.

## How each case gets its pods

| Action | Pods | Cost |
|---|---|---|
| Delete a Pod | its own labels, off the row | no extra call |
| Delete or scale a workload | its `spec.selector` | 1 pod list + 1 budget list |
| Drain | every pod on the node | 1 pod list + 1 cluster-wide budget list |
| Bulk delete | one pod list per distinct namespace | 2 calls for a 50-row selection, not 50 |

A list row carries no labels and no spec, so bulk cannot resolve workload rows without one call per row. Those are reported as `N rows not counted` rather than silently dropped from the total.

Scale fetches once when the overlay opens and recomputes as you type, so the line follows the number without a call per digit.

## Reuse

The advisor already had a correct `policy/v1` selector matcher. It moved to `internal/k8s` as `PDBSelectorMatches` and the advisor now calls it, so the dialog and the security findings cannot disagree. The separate PDB read in `orphan_kinds.go` was deliberately left alone: it distinguishes "no selector" from "invalid selector" to report distinct reasons, which a boolean helper cannot express.

`ConfirmNote` serves both the confirm and the input overlay, so one wording drives both. TASK-862 appends its dependent counts to the same slice.

## Review

Two reviewers, two rounds each, both ended at PASS. What they caught, all fixed here:

- A panic. `scaleBlastRadius` sliced `pods[:len-target]` unclamped; a negative target read past the slice. Only the digit-only key filter kept one out.
- A stuck spinner. `loadScaleBlastRadius` returned no command when the object had no `spec.selector`, so the overlay read "checking disruption budgets" forever. Every loader now answers.
- A retained fetch. The first cleanup fix patched the per-overlay handlers, but Ctrl+C, the toggle key, and a tab switch never reach them. The reset now sits in `closeCurrentOverlay` and `loadTab`.
- Two allocation fixes on the per-keystroke recompute path.

Measured, not assumed: bulk stays bounded by namespace count on a 50-row multi-namespace selection; drain's lists are server-side filtered; the per-keystroke recompute at 500 pods and 50 budgets is roughly 25k cheap comparisons, sub-millisecond. Every fetch runs in the scheduler's worker pool, off the event loop.

## Test plan

- [x] 60 new tests. Budget math covers no budget, zero replicas, nil and empty selectors, a budget in another namespace, and a shortfall that must stay negative
- [x] Full `go test -race ./...` clean, `golangci-lint` 0 issues, `go vet` clean
- [x] `internal/app/app.go` still exactly 800 lines, its revive cap
- [x] Eleven existing tests retargeted: Delete, Drain, Scale, and bulk Delete asserted a nil `Cmd` and now all return a loader. Each still asserts the overlay opens

## Not covered

No test drives a real cluster. The pod and budget fetchers are covered against the fake clientset, which ignores `FieldSelector` and `LabelSelector`, so the production code re-filters on the client and the tests prove that re-filter rather than the server's behavior.


janosmiko:
@coderabbitai review

janosmiko:
@coderabbitai review

PR #593, opened 2026-08-09
fix: open a restored session on its saved view

janosmiko:
Opening lfk on a saved session showed two intermediate states before settling. Both are gone.

## What you saw

1. The explorer opened on the resource-type browser, then jumped to the saved view about a second later.
2. The list then reshuffled under the cursor, which moved from row 3 to row 4.

Both reproduced on a `WorkflowTemplates` view in `dev-envs`, where every row is named `backup-db`.

## Cause 1: the saved type could not be resolved yet

`WorkflowTemplate` is a CRD, so it has no match in the built-in seed list. `restoreSingleTabSession` could not resolve it, parked on the resource-type browser, and navigated only once live discovery answered.

`NewModel` now reads the on-disk discovery snapshot for the one context the session opens, so the saved type resolves before the first frame.

Only the active tab's context is read. Walking every context costs one `clientcmd.ClientConfig()` call each and hangs startup on kubeconfigs with thousands of contexts, which is why the full preload stays async. A test pins that background tabs keep waiting for it.

## Cause 2: the first list was never sorted

`sortMiddleItems` bailed whenever `ui.ActiveSortableColumns` was empty. That slice is filled during a middle-table render, so the first `resourcesLoadedMsg` after launch could never sort. The list painted in apiserver order and the next watch tick sorted it, moving every row that shares a primary key with another.

The sort needs only the column name, which the model already holds, so the render-global check is gone.

## Also: a keystroke during the restore is no longer undone

The saved row still lands when the list arrives, and `restoreCursorAfterLoad` used to apply it regardless of what the user did in between. A key or click during that gap now drops the saved cursor. The guard is scoped to the restore, because `pendingTarget` also carries bookmark jumps and orphan drill-downs.

## Known gap

The first launch after quitting on a CRD view of a cluster lfk has never cached still passes through the type browser once. No snapshot exists yet. Every later start is clean.

## Test plan

- [x] 20 new tests. The sort pair reproduces the reorder before the fix: load 1 `[mget-reports-dev a2community-stage emersys-demo a2community-dev]`, load 2 `[a2community-dev a2community-stage emersys-demo mget-reports-dev]`
- [x] `go build ./...`
- [x] `go test -race ./...`
- [x] `go vet ./...`
- [x] `gofumpt -l internal/` clean
- [x] `golangci-lint run ./...` 0 issues
- [x] Confirmed against a live cluster: opens straight on the saved WorkflowTemplates row, no jump, no reshuffle


janosmiko:
Addressed in 7e0732b5.

**Deferred restore not abandoned on terminal discovery (Major): fixed, and the bug was worse than described.**

Neither terminal path cleared the parked state, and the parked type is matched against whatever context the user is on when discovery next answers. So after discovery failed on one cluster, opening a different cluster resumed the dead restore *there* and dragged its filter along. A regression test pins it:

```
TestDeferredRestore_FailedDiscoveryCannotResumeInAnotherContext
```

Both paths now call `dropDeferredSessionRestore`. I did not take the suggested patch as written: it routes through `finishSessionRestoreForContext`, which on the success path runs after the awaiting field has already been cleared at the top of the branch, so the armed `pendingSessionList` would have survived. The clears sit at the two terminal points instead.

**Assert the complete curated order (Minor): fixed.** `sortAscending` is now explicit and the test compares all four namespaces against the input order.

**Redact discovery failure fields (Minor): skipped.** That log line is pre-existing and untouched by this PR; only the comment above it is new. A kubeconfig context name and a discovery error class are not credentials, and this log shape is used throughout the package, so changing it here alone would be inconsistent adjacent-code churn. Worth a separate pass over the package's logging if you want it.

4 new tests. `go build`, `go test -race ./...`, `go vet`, `gofumpt` and `golangci-lint` all clean.

janosmiko:
Addressed in e61a3e1f. Assertion added; it passes, so the namespace was already carried correctly and this closes a coverage gap rather than a defect. Full suite, -race, vet and lint clean.

janosmiko:
@coderabbitai review

PR #594, opened 2026-08-10
feat: add a built-in demo cluster behind --demo (TASK-865)

janosmiko:
Adds `lfk --demo`: an in-memory fake cluster, so lfk starts with no kubeconfig and no real cluster. Closes TASK-865.

Reuses the existing fake-client seam (`testClientset` / `testDynClient`, now `injected*`) rather than adding a `KubeBackend` interface in front of a `Client` that exposes 150+ methods.

## What it does

- `--demo` builds a fake cluster: 2 nodes, 4 namespaces, a healthy Deployment, a CrashLoopBackOff pod, a failed Job, Services, ConfigMaps, Events, and metrics.
- A 3s ticker mutates the object tracker, so restart counts climb and Warning Events arrive while you watch.
- A DEMO badge marks the session. `NewClient` can never select the demo backend; a test asserts it.
- 32 `exec.LookPath("kubectl")` sites now route through one `k8s.KubectlPath()`. Under `--demo` it returns `os.Executable()`, so call sites re-enter a hidden `__demo-kubectl` serving `logs -f`, `get` and `describe`.
- Generated log lines are JSON access logs that `internal/logagg` detects and groups, so Log Top has real data.

## Not simulated

exec, port-forward, drain, debug, helm and vuln scan are refused with a clear message. Nothing reaches a real cluster, but those actions do not work against the fake either. Documented in README, `docs/usage.md` and `docs/installation.md`.

## Bugs found and fixed on the way

Both got past a fully green `-race` suite and only showed up in a live run:

- The dynamic fake panics on an unregistered ListKind. `fetchCRDPrinterColumns` lists CRDs unconditionally, on a scheduler worker where nothing can recover. Registered a ListKind for every listable resource, not only seeded ones, plus an invariant test. The audit found 9 more, including `resourcequotas`, which would have been the next crash.
- With a real config carrying a `_global` Prometheus block, `--demo` nil-panicked in `queryPrometheusMetric`. Every smoke test used an empty HOME, which is exactly what hid it. Three `ProxyGet` surfaces now guard on demo mode.

## Review findings fixed

- CRITICAL: `client-go`'s `testing.Fake` appends every call to an unbounded slice. Measured 1.52 GB / 12.6M objects at 52,000 ticks; now flat at 2.7 MB / 27K.
- HIGH: Vuln Scan called `exec.LookPath("trivy")` unguarded, so pressing `V` on a demo pod pulled a vulnerability DB over the network. The test proved it by running a real 34s scan before the fix.
- MEDIUM: `safeProxyGetRaw` swallowed real client-go panics with no stack; now logs before returning.
- MEDIUM: `ApplyManifest` validated the name but not the namespace. A bidi override in `metadata.namespace` reached `resourceTitleLabel`, shared by six views.

## Test plan

- [x] `go build ./...`
- [x] `go test -count=1 -race ./...` (25 packages)
- [x] `make lint` (0 issues)
- [x] `make e2e` (new pty-driven harness; teatest pins bubbletea v1, incompatible with this repo's v2)
- [x] Live tmux run, empty HOME and real HOME, no panic
- [ ] Manual UAT per the guide in TASK-865

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: setup lint lint-fix test coverage fuzz build generate-themes sonar bump-version refresh-vendor-hash release goreleaser-check
+.PHONY: setup lint lint-fix test e2e coverage fuzz build generate-themes sonar bump-version refresh-vendor-hash release goreleaser-check
 
 setup:
 	git config core.hooksPath .githooks
@@ -65,6 +65,10 @@ lint-fix:
 
 test:
 	go test ./...
+
+e2e: ## Run e2e tests that drive the built lfk binary under a pty (--demo mode)
+	golangci-lint run --build-tags e2e ./e2e/...
+	go test -tags e2e -count=1 -timeout 2m ./e2e/...
 
 coverage: ## Run tests with coverage report
 	go test ./... -coverprofile=coverage.out
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ go install github.com/janosmiko/lfk@latest
 nix run github:janosmiko/lfk
 ```
 
-`kubectl` is required and must be configured. `helm` and `trivy` are optional (Helm management, image vulnerability scanning).
+`kubectl` is required and must be configured. `helm` and `trivy` are optional (Helm management, image vulnerability scanning). No cluster yet? `lfk --demo` runs against a built-in fake cluster instead.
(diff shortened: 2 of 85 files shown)


janosmiko:
@coderabbitai review

PR #595, opened 2026-08-10
fix: sanitize terminal escapes in describe, YAML, event and editor sinks (TASK-873)

janosmiko:
Closes the terminal-escape gaps left by the existing sanitising, and fixes a blind spot in the sanitiser itself. Closes TASK-873.

## The sanitiser was blind to half the control range

`sanitizeLogLine` only checked `c < 0x20` and `c == 0x7F`, so C1 (0x80-0x9F) passed through untouched. An 8-bit OSC-52 clipboard write (`\x9d52;c;<base64>\x9c`) and an 8-bit CSI (`\x9b31m`) survived byte for byte, reaching any terminal that honours 8-bit C1 (VTE, Linux console). This predates the branch.

The fix is decode-aware, and that detail matters: `sanitizeLogLine` works on bytes, and UTF-8 continuation bytes also live in 0x80-0xBF. A byte-level range test would have corrupted every accented, CJK, emoji and box-drawing character in the log viewer while still passing an injection test. Only the *decoded* rune is checked. `TestSanitizeLogLine_NonASCIIRoundTrip` is the guard.

Verified against overlong encodings, truncated multibyte sequences, a C1 following a valid multibyte character, and a C1 at end of string.

## Sinks closed

| Sink | Was |
|---|---|
| Describe viewer (both render paths) | no sanitising at all |
| YAML view | no sanitising |
| Event viewer | `Reason`/`Message`/`Source`, later also `Type` |
| `SingleLineCell` (Secret, ConfigMap, Label editors) | stripped only `\n`, `\r`, `\t` |
| `addLogEntry` error-log overlay | raw command output and `err.Error()` |
| Pod prefix, and 7 titles via `resourceTitleLabel` | raw kind, namespace, name, container |

`SingleLineCell` and `addLogEntry` were fixed at the source, so all three editors and all 78 `addLogEntry` call sites are covered without per-site edits.

## Describe sanitising moved off the render path

`viewDescribe` is a View function, so it re-sanitised the whole buffer every frame. Sanitising now happens once in the five producers.

| Lines | Before | After |
|---|---|---|
| 20,000 | 2.45 ms, 22,470 allocs | 494 µs, 2,468 allocs |

Allocations are now independent of buffer size.

## What this does not claim

This closes the sinks found across three review rounds. It does not prove no sink remains. `Event.Type` was printed raw in the same `fmt.Sprintf` where three sibling fields had just been sanitised, so the failure is field-level, not function-level, and neither a lint rule nor a "does this function sanitise" test would catch it. The structural fix is a tainted-string type around cluster-sourced fields, forcing an explicit unwrap — a sizeable refactor, deliberately not attempted here.

## Test plan

- [x] `go build ./...`
- [x] `go test -count=1 -race ./...` (23 packages)
- [x] `make lint` (0 issues)
- [x] Injection cases per sink: 8-bit OSC-52, 8-bit CSI, bidi override
- [x] SGR colour and tab alignment survive the describe path
- [x] Non-ASCII round-trip: accented Latin, CJK, emoji, box-drawing
- [x] `BenchmarkViewDescribe` before and after
- [ ] Manual UAT per the guide in TASK-873

### DIFF
--- a/internal/app/breadcrumb_viewers_test.go
+++ b/internal/app/breadcrumb_viewers_test.go
@@ -21,6 +21,28 @@ func TestResourceTitleLabel(t *testing.T) {
 	assert.Equal(t, "x", resourceTitleLabel("", "", "x"))
 }
 
+// resourceTitleLabel feeds seven fullscreen view titles (logs, describe,
+// exec, object explorer, YAML, event viewer, logtop) with kind/namespace/name
+// taken raw from cluster-controlled resources. A hostile name must not be
+// able to reorder or hijack the title bar; fixing it once here covers every
+// call site.
+func TestResourceTitleLabel_SanitizesTerminalEscapes(t *testing.T) {
+	label := resourceTitleLabel("Pod", "default", "evil\x1b[2Jpod")
+	assert.NotContains(t, label, "\x1b")
+	assert.Equal(t, "Pod default/evil[2Jpod", label)
+
+	label = resourceTitleLabel("Pod", "default", "evil\u202epod")
+	assert.NotContains(t, label, "\u202e", "bidi override must not survive")
+
+	label = resourceTitleLabel("Pod", "default", "evil\x1b]52;c;ZXZpbA==\x07pod")
+	assert.NotContains(t, label, "\x1b")
+	assert.NotContains(t, label, "\x07")
+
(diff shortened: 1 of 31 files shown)


janosmiko:
@coderabbitai review

PR #600, opened 2026-08-10
feat: hide actions the current user is not allowed to run (TASK-872)

janosmiko:
## Summary

The action menu offered every action whatever the user's RBAC. A user without delete rights pressed Delete and learned the answer from an API error after the confirm dialog.

A bulk `SelfSubjectAccessReview` pass now runs when a resource list loads, off the key path, and the menu drops the entries the cluster would refuse. Read-only mode and RBAC answer through one predicate, so the menu that hides an entry and the dispatcher that refuses it cannot disagree.

Covers Pod, Deployment, StatefulSet, DaemonSet and ReplicaSet. Every other kind keeps the read-only gate alone.

## How it works

- One pass per context, namespace and kind. The second visit reads the cache and makes no call.
- Each kind asks only what its menu offers: `delete` and `patch` on the object, `update` on `<resource>/scale` for Scale, and the pod-level questions behind Logs, Exec, Attach, Port Forward and Debug Pod.
- The cache key carries the kind, so a Pod verdict never answers for a Deployment menu.
- Bounded: 32 queries per pass at most, 8 in flight, deduplicated, scheduled at low priority.

## Fail open

A hidden action that would have worked is worse than an action that fails, and an aggregated or webhook authorizer can overrule a denial. Every action stays visible when:

- the review errors, times out, or the scheduler drops it (the in-flight marker expires after 30s so the next visit retries)
- the action has no mapped verb, or the kind has no verb map
- the list is all-namespaces, multi-namespace, or union: none of them has one scope a single pass could speak for
- the kubeconfig was re-read, which drops the cache since a context name can then point at another cluster or user

## Test plan

- [x] `go build ./...`
- [x] `go test ./...`
- [x] `go test -race ./internal/app/ ./internal/k8s/`
- [x] `golangci-lint run ./internal/...` (0 issues)
- [x] 32 tests for the feature: verb map against each kind's menu, scope isolation across contexts, namespaces and kinds, every fail-open path, the dropped-pass retry, and the menu dropping refused entries
- [ ] Manual: open a Pod list in a namespace without `delete` on pods, press `x`, confirm Delete and Force Delete are gone; repeat on a Deployment list without `update` on `deployments/scale` and confirm Scale is gone

## Review

Three review passes applied: a Go review (one HIGH: the in-flight marker never cleared when the scheduler dropped a superseded pass), a security review (two MEDIUM: all-namespaces reviewed a stale namespace, union lists covered one member cluster; one LOW: stale cache after a kubeconfig reload), and a Go review of the workload extension (approved, no CRITICAL or HIGH).

Closes TASK-872.


janosmiko:
Addressed in c56b16b3.

**Fixed**
- *Generated force-action labels*: `Force Finalize` is now in the verb map for every covered kind, mapped to `patch`, which is what removing finalizers sends. It shares the key `Edit` and `Restart` already ask for, so it costs no extra review. `Force Delete` was already mapped for Pods; workload kinds are not force-deletable, so their deleting rows generate `Force Finalize` instead.
- *Bulk actions*: `executeBulkAction` and the bulk menu now consult the gate. It answers only when every selected row sits in one context and namespace; a mixed selection has no single verdict, so the API server decides row by row. Read-only keeps its all-or-nothing behavior, which RBAC deliberately does not copy.
- *Cap test*: every query is now unique, and the assertion is exact rather than `LessOrEqual`.

**Skipped**
- *Cached verdicts in ambiguous list scopes*: not a defect. `deniedByRBAC` reads `m.actionCtx.namespace`, which is the selected row's own namespace, and the cached verdict was obtained by reviewing exactly that namespace and kind. It stays correct however wide the surrounding list is; failing open there would discard a valid answer. What was wrong was the doc sentence claiming every action stays visible in those lists - corrected to say no review is *sent*, while a row whose namespace was reviewed earlier still uses that verdict.

janosmiko:
Both remaining findings fixed in 3c4522df.

- *`Group` missing from `PermissionQuery.Key`*: real collision. The key now carries the group in the `resource.group` form kubectl uses, so `delete:widgets` and `delete:widgets.example.com` stay apart. Two tests cover it: one on the key, one driving `CheckPermissions` with a core and a CRD resource of the same name and opposite verdicts.
- *README scope*: now states context, namespace and kind.

PR #609, opened 2026-08-11
feat: add an Undeliverable view for resources stuck waiting (TASK-855)

janosmiko:
A cluster-wide list of resources that want to reach a state and cannot, each row carrying the reason from its own API object. Mirror of the orphan overview (`Shift+Z`), which finds resources nobody references.

Opens with `V`, or `:undeliverable` / `:stuck`, and sits under the Dashboards section beside Cluster and Monitoring.

## Detectors

Every reason comes from a first-class field, a status condition, or a verbatim `Reason: Message` off a matched Event. No heuristics, no string-matching on free text - a guessed reason is worse than no row.

| Resource | Reason source |
|---|---|
| Pending Pod | `FailedScheduling` Event, `PodScheduled` condition as fallback |
| Pending PVC | binding Events, or an unset `storageClassName` |
| Service | ready endpoint count across its EndpointSlices |
| Ingress | empty `status.loadBalancer` |
| Any object terminating | the finalizers named on it |

## Also in this PR

- **A real bug found while wiring**: the inflight slot was cleared *after* the context check, so a scan landing for a cluster the user had left never freed the slot and deadlocked every later scan on the spinner. Covered by a test.
- `model.IsDashboardPseudoItem` replaces a hand-written `"__overview__" || "__monitoring__"` pair at seven sites. Six such pairs remain, all genuinely dashboard-specific behaviour - Undeliverable is a Dashboards entry but has no preview pane.
- Four ceiling problems had to be cleared first: three switches were exactly at the gocyclo 30 limit and `app.go` at exactly the 800-line cap. Each switch's fallthrough moved to a tail handler, and the finalizer-search fields moved into an embedded `finalizerSearchState`, matching the existing `columnToggleState` / `canIState` pattern. `app.go` is now 795.

## Security

Every rendered value here is cluster-sourced, so an attacker who can create objects in a watched namespace controls it. A security review of the diff found the overlay path clean - Kind, Namespace, Name and Reason all go through `SanitizeTerminalText` before any `Truncate`, so the sanitize-before-measure ordering holds.

It found one leak, now fixed in `fix: sanitize the scan error on its way to the status bar`: the partial-scan error reached `setStatusMessage` raw, and that sink strips only `\n` and `\r`. An OSC-52 payload survived to the far end with its introducer and BEL intact. The fix sanitizes before concatenating, matching what the overlay subtitle already did.

The underlying cause is broader than this feature and is filed as **TASK-884**: `setStatusMessage` never sanitizes, so any caller interpolating a cluster string leaks. `handleOrphansLoaded` has the identical bug today. Fixing it at the sink covers every caller at once, including unaudited ones, which is why it is not a drive-by here.

## Docs

README features, `docs/usage.md`, `docs/keybindings.md`, `docs/views-and-overlays.md`, and the in-app help.

Separately filed as **TASK-883**: README says `Shift+Z` for the orphan overview while two docs files say `Shift+O`. The binding is `Z`, so those two are stale. Untouched here to keep this change in scope.

## Verification

`gofmt`, `go build`, `go vet`, `golangci-lint` 0 issues, `go test -count=1 -race ./...` exit 0 with a per-package `ok` across all 25. Per-detector table tests plus 18 overlay tests; a reviewer checked each for vacuity and found none.

## Known gap

A context switch does not cancel an in-flight scan. The generation and context guards keep the data correct, but one wasted read-only list round happens. Cancelling properly needs edits at the six `invalidateOrphanCacheForContext` call sites, which looked like churn for a read-only scan.

Closes TASK-855.


janosmiko:
All six findings addressed across `feecb72e`, `59c66f59`, `32a8d7d6`, `ed9154ff`.

## MAJOR - unreadable EndpointSlice list read as an empty one

Right, and the cause sits deeper than the line you flagged. client-go's `List()` allocates its result struct before it can fail, so it **never** returns a nil pointer - which means the `safe*Items` nil guards this code relied on were dead code. A failed list was indistinguishable from an empty one.

Fixed at the source in `fetchUndeliverableLists` by nilling the pointer on error, then skipping the Service detector when `lists.slices == nil`. The partial-scan error still surfaces so the overlay explains the omission.

Verified independently, not just reported. On the unfixed code:

```
Error:    Should be empty, but was [{web live Service no EndpointSlice endpoints}]
Test:     TestDetectUndeliverable_EndpointSliceListFailureSkipsServices
Messages: a failed EndpointSlice list must not read as every Service having no endpoints
```

Green after. That is exactly the false positive you predicted.

I asked for the other detectors to be audited for the same shape. Result: Pods, PVCs, Ingresses and Terminating each depend only on their own primary list, so a failure there omits rows rather than inventing them. A failed Events list makes the Pod and PVC detectors fall back to their own first-class sources - status conditions and `spec.storageClassName` - so no heuristic and no false positive. **Services were the only cross-list dependency**, where one list succeeding and another failing produced a wrong answer instead of a missing one.

`undeliverable_detector.go:88` is not the same bug - that is `latestReason`'s safe fallback path, already correct.

## MAJOR - stale scan leaves the overlay empty

Fixed. `handleUndeliverableLoaded` now restarts the scan when a stale result lands while the overlay is open for a newer context. Test `TestHandleUndeliverableLoaded_RestartsScanWhenOverlayOpenForNewContext`, red beforehand with `Expected value not to be nil`.

## MAJOR - hand-computed cursor window

Fixed. `UndeliverableScrollForCursor` delegates to `ui.VimScrollOff`. Test red beforehand at `UndeliverableScrollForCursor = 5, want 4`.

Worth saying plainly: this convention was in the implementer's brief and two reviewers missed it. Good catch.

## MINOR - rune-count padding

Fixed. Columns pad through `padRight`, which measures with `lipgloss.Width`, applied after sanitize and truncate. Test red beforehand at `row display widths differ: ascii=78 wide=79`.

## MINOR - single-cluster precondition, and README wording

Both fixed. Confirmed `openUndeliverableOverlay` does reject `isUnionSentinel()` before documenting it, in `docs/keybindings.md` and `docs/usage.md`. README:394 now reads "from the resource's API fields, status conditions, or Events."

## Filed separately: the same class in the orphan scanner

Your finding generalises, and the generalisation is worse. The orphan scanner decides an object is unused by checking whether anything references it - a cross-list dependency by definition. A failed list there is collected into a partial-result banner, but the rows are still shown as orphans. So a failed Pod list makes a Pod-only-referenced ConfigMap look orphaned, in a view that offers bulk deletion.

Filed as TASK-886 rather than fixed here, since it changes what a different feature reports. The `safe*Items` dead-code point applies there too.

Gate: `gofmt`, `go build`, `go vet`, `golangci-lint` 0 issues, `go test -count=1 -race ./...` green across 25 packages. `app.go` untouched at 795.

janosmiko:
Rebased onto current `main` and pushed as `12a8d0d0`. Conflicts resolved, now `MERGEABLE`.

## The one real conflict

`docs/views-and-overlays.md` - #611 corrected the orphan overlay row from `Shift+O` to `Shift+Z` while this branch added the Undeliverable row directly beneath it. Both changes wanted, so the resolution keeps the corrected `Shift+Z` and the new row.

Everything else auto-merged, including `docs/keybindings.md` and `docs/usage.md`.

## Semantic overlap with what landed while this was open

A clean textual rebase is not a semantic one, so these three were checked rather than assumed:

- **#610 (TASK-874, tainted strings)** - the build is the check here, and that is the point of that change. `EventInfo`'s fields are now `tainted.String`, so if this feature's detectors read an Event field without unwrapping through a sanitizer, it would not compile. It compiles.
- **#608 (TASK-879, tabs.go split)** - no overlap; this branch does not touch the moved code.
- **#612 (TASK-884, sanitizing at the status-bar sink)** - this branch also sanitizes the scan error at its call site (`internal/app/commands_undeliverable.go:80`). That is now redundant, since the sink sanitizes everything. Left in place deliberately: `SanitizeTerminalText` only deletes matching runes, so re-running it is a no-op, and removing the call would widen this diff for no behavioural gain. Flagging it so it reads as intentional rather than overlooked.

## Verification on the new base

`gofmt` clean, `go build ./...` clean, `go vet ./...` clean, `golangci-lint run ./...` **0 issues**, `go test -count=1 -race ./...` exit 0 across 26 packages.

One note on the lint run, in case anyone reproduces it: the first pass reported 3 issues, one of them citing a file path inside a git worktree that no longer exists. That was a stale golangci-lint cache. After `golangci-lint cache clean` the run is clean.

PR #610, opened 2026-08-11
feat: add a tainted-string type for cluster-sourced fields (TASK-874)

janosmiko:
Terminal escape injection through cluster-controlled strings has recurred four times. TASK-873 closed six sinks across three review rounds, each round finding more. TASK-880 then found ten that TASK-873 had missed, including a field printed raw in the same `fmt.Sprintf` where three sibling fields had just been sanitized. TASK-855 shipped one more this week with the threat spelled out in its brief.

The failure is field-level, so a lint rule or a "does this function sanitize" test passes on the code while it still leaks. This makes the next one a compile error instead.

## The type

New leaf package `internal/tainted`, stdlib imports only, so `k8s`, `model` and `ui` can all use it.

`type String struct{ raw string }` - a struct, not a named string. That distinction is the whole design: `type Tainted string` still satisfies `%s`, still concatenates, and still passes where a `string` is wanted, so it buys nothing. The struct cannot concatenate, cannot compare to a literal, and cannot pass as a `string`.

It implements `fmt.Formatter`, so `%s %v %q %d %x %#v %+v`, width and precision flags, and the zero value all emit `[tainted:unsanitized]` rather than the payload.

Two unwraps, no default, so the choice between the sanitizers is forced:

| Method | Applies |
|---|---|
| `.Line()` | `SanitizeTerminalText` - single-line values |
| `.Body(renderAnsi)` | `StripBidiOverrides` + `SanitizeLogBody` - multi-line bodies |

Non-render needs are served by `Is()`, `IsEmpty()`, `Compare()` and `Contains()`, which is why `.Raw()` has **zero production call sites** - the 6 that exist are all test assertions.

## The sanitizers moved down a layer

The three implementations moved out of `internal/ui` into `tainted`, and `ui` re-exports them, so its existing call sites and tests are untouched. `logviewer.go` 777 to 628, `listsummary.go` 457 to 430.

This was the riskiest part of the change, since roughly 200 existing sinks now reach the sanitizers through a delegation. A security review diffed the relocated implementations against `main` line by line - byte-identical apart from comment placement - then set up a throwaway `main` worktree and pushed six payloads (OSC-52, CSI erase, raw C1, bidi override, tab, SGR) through all three functions on both branches. Byte-for-byte identical output in every case. The re-exports are one-line delegations, not reimplementations.

## Scope

`k8s.EventInfo` and its mirror `ui.EventTimelineEntry` - Type, Reason, Message, Source, InvolvedName, InvolvedKind - converted end to end, from the producer in `client_rbac.go` through both render sinks. Reviewers traced all six fields and found no early unwrap.

Deliberately left out: object names and namespaces, Secret and ConfigMap values, label and annotation values. Those flow through `model.Item` and the whole explorer column pipeline, hundreds of sites. A partial conversion there would leave gaps the type system then blesses, which is worse than the status quo. **They remain plain strings relying on manual sanitizer calls, the same residual risk as before this PR.**

## Proof it works

Removing a real sanitizer call now breaks the build:

```
internal/ui/explorer_resource.go:501:13: cannot use e.Reason (variable of struct type
  tainted.String) as string value in struct literal
internal/app/update_overlays_events.go:58:20: cannot use e.Type (variable of struct type
  tainted.String) as string value in argument to padColumn
```

One shape still compiles, and is documented rather than hidden: assigning the field and passing it straight to a verb, `message := e.Message` then `%s`. It does not leak - it renders `[tainted:unsanitized]` and turns `TestBuildEventTimelineLines` red. A reviewer probed nine more such shapes (map ranged, slice element, `fmt.Errorf`, `errors.New`, struct embedding, assertion to `fmt.Stringer` through `any`, width and precision flags, zero value, `%x`) and every one emits only the marker.

So: compile error where the value crosses into a string parameter, loud marker plus a red test where it goes straight to a verb. Never silent.

## Tests

The two guard tests are the durable part. Both iterate struct fields with `reflect.TypeFor[T]().Fields()` rather than a hardcoded list, so a newly added raw field fails rather than slipping through.

Every test was proven red by name before being made green, and a reviewer independently re-verified three by reverting the method each covers.

## Verification

`gofmt`, `go build`, `go vet`, `golangci-lint` 0 issues, `go test -count=1 -race ./...` exit 0 with 26 of 26 packages `ok`. Largest touched file 778 of the 800-line cap.

## Follow-ups, not fixed here

- `json.Marshal` on a struct holding a `tainted.String` silently drops the field, since `raw` is unexported. No leak and no call site does this today, but it is a trap worth knowing.
- `InvolvedName` and `InvolvedKind` are wrapped and carried through, then never rendered. Dead fields.
- **TASK-884** is the related one worth attention: `setStatusMessage` never sanitizes, so any caller interpolating a cluster string leaks into the status bar. `handleOrphansLoaded` has that bug today.

Closes TASK-874.

### DIFF
--- a/internal/app/recompose_theme_test.go
+++ b/internal/app/recompose_theme_test.go
@@ -12,6 +12,8 @@ import (
 	"github.com/janosmiko/lfk/internal/k8s"
 	"github.com/janosmiko/lfk/internal/model"
 	"github.com/janosmiko/lfk/internal/ui"
+
+	"github.com/janosmiko/lfk/internal/tainted"
 )
 
 // themeA and themeB differ only in their Base background, which is enough for
@@ -45,7 +47,7 @@ func TestRecomposeThemedContentRerendersCachedPreviews(t *testing.T) {
 			memUsed: 256, memReq: 512, memLim: 1024,
 		},
 		previewEventsData: []ui.EventTimelineEntry{
-			{Timestamp: time.Time{}, Type: "Warning", Reason: "BackOff", Message: "boom", InvolvedName: "p", InvolvedKind: "Pod"},
+			{Timestamp: time.Time{}, Type: tainted.Wrap("Warning"), Reason: tainted.Wrap("BackOff"), Message: tainted.Wrap("boom"), InvolvedName: tainted.Wrap("p"), InvolvedKind: tainted.Wrap("Pod")},
 		},
 		monitoringData: map[string]monitoringData{
 			"ctx": {alerts: []k8s.AlertInfo{{Name: "HighCPU", State: "firing", Severity: "critical"}}},
--- a/internal/app/renderoverlay_test.go
+++ b/internal/app/renderoverlay_test.go
@@ -10,6 +10,8 @@ import (
 	"github.com/janosmiko/lfk/internal/k8s"
 	"github.com/janosmiko/lfk/internal/model"
 	"github.com/janosmiko/lfk/internal/ui"
+
+	"github.com/janosmiko/lfk/internal/tainted"
 )
 
 // baseOverlayModel returns a minimal Model for overlay rendering tests.
@@ -309,13 +311,13 @@ func TestRenderOverlayEventTimeline(t *testing.T) {
 	m.eventTimelineData = []k8s.EventInfo{
 		{
 			Timestamp:    time.Now(),
-			Type:         "Normal",
-			Reason:       "Created",
-			Message:      "Created container nginx",
-			Source:       "kubelet",
+			Type:         tainted.Wrap("Normal"),
+			Reason:       tainted.Wrap("Created"),
+			Message:      tainted.Wrap("Created container nginx"),
+			Source:       tainted.Wrap("kubelet"),
 			Count:        1,
-			InvolvedName: "my-pod",
-			InvolvedKind: "Pod",
+			InvolvedName: tainted.Wrap("my-pod"),
+			InvolvedKind: tainted.Wrap("Pod"),
 		},
 	}
 	m.actionCtx = actionContext{name: "my-pod"}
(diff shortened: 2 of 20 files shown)


janosmiko:
Both findings addressed in `4bfab6cd`.

## Private CSI markers (Major) - fixed, and it was a live hole

You were right, and it is worse than the parameter range suggests. Verified before and after with a probe through `SanitizeLogBody(s, true)`:

| Payload | Before | After |
|---|---|---|
| `\x1b[>4;2m` XTMODKEYS | `pre\x1b[>4;2mpost` **intact** | ESC replaced |
| `\x1b[?1049m` | intact | ESC replaced |
| `\x1b[<3m` | intact | ESC replaced |
| `\x1b[1 m` intermediate | intact | ESC replaced |

The parameter scan now accepts only `0-9`, `;` and `:` (`internal/tainted/sanitize.go:199`), rejecting the private markers and the `0x20-0x2F` intermediates.

Legitimate colour still survives, which was the risk in tightening this - 8-colour, 256-colour, reset, combined attributes, and truecolour in **both** the semicolon and colon forms all pass unchanged. `TestSGRParameterBytesKeepLegitimateColour` guards that; `TestSGRParameterBytesRejectPrivateMarkersAndIntermediates` guards the rejection and was proven red on the old parser.

Worth flagging for the record: **this is pre-existing, not introduced here.** The identical bytes are on `main` at `internal/ui/logviewer.go:763-768`, so it ships today. It is filed as TASK-885 and fixed in this PR because this PR relocates the file into `internal/tainted`, and shipping the move with a known hole in it would be worse.

## `len(e.reason)` (Minor) - fixed, and this PR did touch it

Checked before deciding: `git diff origin/main...HEAD` shows the hunk rewrote that exact line, so it is in scope here rather than pre-existing.

Fixing the measurement alone was not enough. With `lipgloss.Width` feeding the old byte-slicing `truncateStr`, a multibyte reason still got cut mid-rune - reproduced with a literal `\xe6` half-rune in the output. The render loop now uses the width-safe `Truncate` and `padRight` helpers. `TestRenderPreviewEvents/multibyte_reason_column_stays_within_width_with_valid_UTF-8` covers it, proven red beforehand.

Gate: `gofmt`, `go build`, `go vet`, `golangci-lint` 0 issues, `go test -count=1 -race ./...` green across 26 packages.

PR #615, opened 2026-08-11
fix: fail loudly on a marshalled tainted.String, drop dead event fields (TASK-887)

janosmiko:
Three loose ends a review of #610 found and deliberately left alone there, so as not to widen that diff.

## `json.Marshal` no longer drops a tainted field in silence

`json.Marshal` on a struct holding a `tainted.String` produced `{"Message":{}}` - the field vanished, because `raw` is unexported and the type had no marshaller. No leak, but a quiet failure that costs someone an afternoon later.

Two designs were on the table. **Chosen: fail loudly.** `MarshalJSON` now returns an error rather than serializing.

The rejected option was a `MarshalJSON` that returns `Line()`. It looks friendlier and is worse: it would become a **second unwrap path that bypasses the deliberate choice between `.Line()` and `.Body()`**, which is the guarantee the whole type exists to provide. An error at the call site is the honest answer - anyone who genuinely wants this text serialized should pick a sanitizer explicitly.

`UnmarshalJSON` needed nothing: `encoding/json` already errors decoding a JSON string into the struct (`cannot unmarshal string into Go struct field ... of type tainted.String`). A test for that would have passed on the unfixed code, so it was deleted rather than kept as decoration.

Checked every `json.Marshal` and `json.Unmarshal` call site in the repo first: nothing marshals `EventInfo` or `EventTimelineEntry` today, so this changes no current behaviour.

## Dead wrapped fields removed

`InvolvedName` and `InvolvedKind` were wrapped as `tainted.String` at the producer, copied field-to-field through `update_metrics_msgs.go`, and never rendered anywhere.

`git log -p` shows why: they were only ever used as local variables for the exact and prefix match filter inside `GetResourceEvents`. The struct fields duplicating them were never read - the timeline renderer only draws Timestamp, Type, Reason, Message and Source. Deleted from both structs and the copy site.

Dead fields that look load-bearing are worse than no fields, particularly when they are wrapped in a type whose purpose is to signal "this needs careful handling".

The two reflection guard tests, `TestEventInfoHasNoRawStringFields` and `TestEventTimelineEntryHasNoRawStringFields`, still pass after the deletion - confirmed rather than assumed, since they iterate struct fields.

## Not in this PR

The third item on TASK-887, converting the six remaining hand-written `"__overview__" || "__monitoring__"` pairs to `model.IsDashboardPseudoItem`, touches files #609 also changed. It waits for that to land.

## Tests

`TestMarshalJSONFailsLoudly`, proven red on the unfixed code: `Error: An error is expected but got nil.`

## Verification

Closes part of TASK-887.


janosmiko:
Not acting on this one, for two reasons.

**This PR does not touch that conversion.** The only change on that line is gofmt realignment - removing the `InvolvedName` and `InvolvedKind` fields shortened the longest key in the struct literal:

```diff
-	Count:        int32(countVal),
+	Count:     int32(countVal),
```

`int32(countVal)` is pre-existing and unchanged. The realignment is what pulled it into the diff.

**The out-of-range value cannot occur.** `Event.count` is declared `int32` in `k8s.io/api/core/v1` (`types.go`, module `k8s.io/api@v0.35.2`). lfk reads it through `NestedInt64` off an unstructured object, but every source of that object is int32-constrained already:

- the apiserver, which decodes and validates against the typed schema, so a larger count never survives admission
- demo mode, where events come from `podWarningEvent(..., count int32, ...)` in `internal/k8s/demo/events.go` - the parameter is already `int32`

Both checked, not assumed. So a range guard would be dead code and a test for a count below `math.MinInt32` would assert a state Kubernetes cannot produce. This repo's guidance is explicit about not adding error handling for impossible scenarios.

What would change the answer: lfk reading Events from something that is not the apiserver and not its own typed demo data - a cached file, a fixture, an external feed. There is no such path today. If one is ever added, the conversion needs the guard at that point, and this comment is a good record of why.

PR #616, opened 2026-08-11
fix: stop a bare sync status posing as Application child health (TASK-888)

janosmiko:
Drilling into an ArgoCD Application listed its child resources with a bare green `Synced`, so a Degraded child read as healthy.

## What was actually wrong

Not the code. `argoStatusResourcesToItems` already read `res["health"]["status"]` and rendered `health/sync`. The data was not there.

Verified directly against a live cluster:

- `status.resourceHealthSource` is `appTree`, and with that setting ArgoCD **does not persist per-resource health into the Application CR at all**. Every entry carries only `group`, `kind`, `name`, `namespace`, `status`, `version`.
- Confirmed on a Degraded app *and* a Healthy one, so it is cluster-wide, not a symptom of the degradation.
- The application-controller has no health-persistence flag set, so this is the **ArgoCD default**. Any default install behaves this way, which makes it the common case rather than an edge case.
- The web UI reads child health from the app-tree cache via the ArgoCD API. lfk does not use that API, and adding it was explicitly out of scope here.

So the old output was worse than showing nothing: it presented a sync status as though it were a full verdict.

## The change

| Health in the CR | Rendered |
|---|---|
| absent | `Unknown/Synced` |
| present | `Degraded/Synced` - unchanged |

`Unknown` is a real ArgoCD health status, not an invented placeholder, so it is the honest label rather than a `-` or `?`.

**Colour was the actual harm and is handled through the existing mechanism.** `Synced` alone lands in the healthy-green `sevRunning` bucket via `statusSeverity` / `phraseSeverityWords`. Adding `"unknown": sevProgressing` to that word map puts `Unknown/Synced` in the amber progressing bucket instead, so it can never read as green. No new colour path was invented. (`internal/ui/styles.go` is at exactly the 800-line cap, so this extends an existing map line rather than adding one.)

**The reason is discoverable.** A `Health Message` column carries `ArgoCD is not persisting per-resource health on this cluster (resourceHealthSource: appTree)`, reusing the details-panel `messageKeys` bucket that already renders that key. It is added only when health is absent, never otherwise, and it stays out of the row.

## Tests

Three proven red on the unfixed code:

- `TestArgoStatusResourcesToItems_MissingHealth_IsLabeledUnknown` - `expected: "Unknown/Synced" actual: "Synced"`
- `TestArgoStatusResourcesToItems_MissingHealth_ExplainsWhy` - no `Health Message` column found
- `TestStatusSeverity_UnknownHealthIsNotColoredHealthy` - `expected: 6 actual: 0` on the RGBA foreground compare

Plus `TestArgoStatusResourcesToItems_PresentHealth_Unchanged`, which passes either way by design - it exists to catch a regression in the working path, not to prove this fix.

A pre-existing test had an unexercised no-health fixture; it now asserts `Unknown/Synced`.

**Also verified against the real thing.** The captured `status.resources` from the cluster that produced the bug was fed through the fixed function: all five ExternalSecrets render `Unknown/Synced` with the message attached.

## Docs

One line on the README's Applications bullet noting that child health needs `resourceHealthSource: resourceHealth` on the ArgoCD side. `docs/usage.md` and `docs/views-and-overlays.md` do not describe this drill-down, so they were left alone.

## Not fixed here

Two things found in the same investigation, both out of scope:

- The app's destination is a **different cluster** (`spec.destination.name: tixx-baseus-dev-argocd-poc`). lfk is connected to the management cluster, where the `ExternalSecret` CRD is not installed - that is the cause of the `Warning: unknown resource type: ExternalSecret` in the drill-down.
- Getting real child health needs either the destination cluster's own objects or the ArgoCD API. Both are larger than this change.

## Verification

Closes TASK-888.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ Namespaces are **not** a navigation level. The current namespace is shown in the
 - **StatefulSets / DaemonSets / Jobs** show their Pods directly
 - **CronJobs** show their Jobs
 - **Services** show Pods matching the service selector
-- **ArgoCD Applications** show managed resources (from status or label discovery)
+- **ArgoCD Applications** show managed resources (from status or label discovery); child health shows as `Unknown` unless the cluster sets `resourceHealthSource: resourceHealth`
 - **Helm Releases** show managed resources (via `app.kubernetes.io/instance` label)
 - **Pods** show their Containers
 - **ConfigMaps / Secrets / Ingresses / PVCs** show details preview (no children)
--- a/internal/k8s/client_fakeclient_extra_test.go
+++ b/internal/k8s/client_fakeclient_extra_test.go
@@ -1203,10 +1203,15 @@ func TestGetArgoManagedResources_WithStatusResources(t *testing.T) {
 	items, err := c.getArgoManagedResources(t.Context(), dc, "", "argocd", "my-app")
 	require.NoError(t, err)
 	assert.Len(t, items, 2)
-	// First should have combined status.
 	for _, item := range items {
-		if item.Name == "my-deploy" {
+		switch item.Name {
+		case "my-deploy":
+			// Health present: combined status, unchanged.
 			assert.Equal(t, "Healthy/Synced", item.Status)
+		case "my-svc":
+			// Health absent (ArgoCD's appTree default): must not read as
+			// healthy just because it's a bare "Synced".
+			assert.Equal(t, "Unknown/Synced", item.Status)
 		}
 	}
 }
--- a/internal/k8s/gitops_argo.go
+++ b/internal/k8s/gitops_argo.go
@@ -67,6 +67,14 @@ func argoStatusResourcesToItems(resources []any) []model.Item {
 			healthStatus, _ = health["status"].(string)
 		}
 
+		// resourceHealthSource: appTree (ArgoCD's default) never writes a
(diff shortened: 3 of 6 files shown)


janosmiko:
Closing this in favour of an ArgoCD-side setting.

The diagnosis stands: `status.resourceHealthSource` is `appTree`, so ArgoCD does not write per-resource health into the Application CR, and there is nothing for lfk to read. But the fix belongs on the ArgoCD side, not here.

The lever was already wired on the cluster, just unset:

```
ConfigMap  argo-cd/argocd-cmd-params-cm
Key        controller.resource.health.persist   (absent -> defaults to false)
```

It feeds `ARGOCD_APPLICATION_CONTROLLER_PERSIST_RESOURCE_HEALTH` on the application-controller. Setting it to `"true"` makes ArgoCD persist health into the CR, `resourceHealthSource` flips to `resourceHealth`, `status.resources[]` gains its `health` key, and the code already on `main` renders `Degraded/Synced` with no change at all.

Reading the children from the destination cluster was the other option and was rejected: lfk should not reach into a cluster the user did not point it at.

Branch left in place in case the `Unknown` rendering is wanted later for default ArgoCD installs, where this is still the out-of-the-box behaviour.

PR #618, opened 2026-08-11
feat: export a live resource as a reusable template (TASK-866)

janosmiko:
Adds an action that strips a live object down to something you can apply somewhere else, and a template directory you can hand-author in.

Press `x` in the action menu on any resource, then pick a destination: clipboard, a file, or your template directory.

## The template directory

`<ConfigDir>/templates/`, one plain Kubernetes manifest per file. The export action writes there, and you can drop files in by hand - **it is the same directory**, deliberately. Adding a template is `cp`, removing one is `rm`. They load into the "Create from Template" picker alongside `BuiltinTemplates()`.

Plain manifests rather than an envelope format, because this is now a user interface: the file name is the template name, the category is `User`, and the description is the manifest's `kind`.

A user template **never silently shadows a built-in**. A user `Pod` and the built-in `Pod` both stay listed, user first, told apart by the `User` category column - otherwise you get the "why is my Pod template different" bug with no way to see why.

A malformed file is skipped and logged, and its siblings still load. You will hand-edit these and get YAML wrong sometimes; one bad file must not empty the picker. A missing directory is normal, not an error.

## The strip list

The point is acceptance criterion #4: **the output has to actually apply**. A manifest the apiserver rejects is worse than useless, because you only find out at `kubectl apply` time.

| Kind | Stripped |
|---|---|
| all | `status`, the server-set `metadata` fields, `last-applied-configuration`, `deployment.kubernetes.io/revision`, embedded pod-template `creationTimestamp` |
| Pod | `spec.nodeName`, `spec.serviceAccount`, the injected `kube-api-access-*` projected volume **and** its volumeMounts |
| Service | `clusterIP`, `clusterIPs`, `ipFamilies`, `ipFamilyPolicy` |
| PVC | `spec.volumeName` and the five binding annotations |
| Job | `spec.selector` and the `controller-uid` / `job-name` labels on the Job and its pod template |
| CronJob | `spec.jobTemplate.metadata` as well as both nested pod templates |

The Job case is the sharp one: **keeping `spec.selector` makes the API reject the create**, so leaving it in would have shipped a template that never applies.

`metadata.namespace` is dropped so `kubectl apply -n` decides.

**Deliberately not stripped**: Service `nodePort`, default tolerations, `spec.priority`. All plausible as author-written, and an over-broad strip silently deletes your work - worse than a leftover field you can see.

## Secrets are redacted

`data` and `stringData` keys are kept with empty values; `type`, labels and annotations survive. A security review found the values passing through byte for byte, which matters here more than elsewhere because all three destinations persist and people commit their config directory to dotfiles.

```yaml
kind: Secret
type: kubernetes.io/basic-auth
data:
  password: ""
  username: ""
```

A template is a shape, not a payload. No opt-in flag for the real values - that would be a config surface, a docs burden and a foot-gun, and `y` already copies the raw object. The status line says the values were redacted, so nobody discovers it by pasting an empty template later.

## Security

Reviewed as a new trust boundary, since this writes files named from cluster data and parses YAML you might paste from the internet. Verified empirically, not by reading:

- The path-traversal matrix is rejected: `../escape`, `..`, `/etc/escape`, `a/../../escape`, `foo/bar`, empty, `.`. `..%2f` stays a literal filename, since Go does not URL-decode.
- **A pre-planted symlink at the target is not followed.** This was initially true only of the template-directory path, which uses `writeFileDurable`'s temp-and-rename. The *file* destination used `O_TRUNC` and did follow a symlink - caught by CodeRabbit after this description was first written, and proved with a symlink over a file containing known content. `writeSecureFile` now delegates to `writeFileDurable`, so all three destinations rename onto the path. Verified after the fix: the target file is unchanged and the symlink itself is replaced.
- A genuinely hostile filename containing OSC-52 loads with the ESC and BEL stripped from the rendered name.
- The save name can only come from a live `metadata.name`, apiserver-constrained to DNS-1123. No other caller reaches the save with a looser string.

The exported manifest itself is left unsanitized on the clipboard, on purpose: it is a functional artifact meant to survive intact, and stripping escapes could corrupt a field you meant to apply. The rendered surface is separately clean.

## Read-only mode

Allowed, per AC #6. `"Export Template"` is simply absent from `mutatingActions`, so the gate never sees it. Nothing here touches the cluster.

## Tests

Per-kind table tests assert **both halves** - the server-set field is gone AND a meaningful field survived. The second half is what catches an over-broad strip.

Every test was proven red before being made green, and the writer **deleted two of its own as vacuous** after finding no mutation could make them fail. A reviewer independently mutation-checked three more, including removing `stripJob`'s `delete(spec, "selector")`, and all three went red.

AC #4 is covered by a strict typed round-trip: `sigsyaml.UnmarshalStrict` into `corev1.Pod`, `Service`, `PVC`, `batchv1.Job`, `appsv1.Deployment`, `corev1.Secret`, plus a sweep asserting no server-assigned *value* survives anywhere in the output.

## Two shared-file edits the linters forced

- `writeFileDurable` lost its `mode` parameter once a fourth call site landed and `unparam` fired. All three pre-existing callers passed `0o600` and the function still pins 0600 internally - verified before accepting.
- One blank line came out of `app.go`, which sits at exactly the 800-line cap.

## Follow-ups filed, not fixed here

- **TASK-890**: two same-named objects in different namespaces clobber each other's saved template. Real, but it has three defensible answers, so it gets its own decision.
- `isSafeTemplateName` accepts a leading-dot name. Unreachable from a real cluster name, since DNS-1123 forbids a leading dot.

## Verification

Closes TASK-866.


janosmiko:
All six addressed in `0f1112a0`.

## Major - the file export followed a symlink

You were right, and it also **corrects a claim in this PR's description**, which I have now edited.

The description said a pre-planted symlink is not followed. That was true of the *template directory* path, which goes through `writeFileDurable`'s temp-and-rename - a security reviewer tested that one and it held. It was **not** true of the *file* destination, which used `writeSecureFile` and `O_TRUNC`. I generalised one result to all three destinations without checking the third.

Proved before fixing: a symlink named `pod_web.template.yaml` pointing at a file containing `ORIGINAL SECRET CONTENT`, written through that exact open call, left the target holding the manifest.

Two things made it worse than the pre-existing use of that helper. The filename is predictable - `templateExportFilename` builds `<kind>_<name>.template.yaml` from the object's kind and name - and it lands in whatever directory lfk was launched from.

**Fixed at the source**, not just on the export path. `writeSecureFile` now delegates to `writeFileDurable`. Its only other caller is `exportResourceToFile`, also an export-to-cwd path with no dependence on truncate-in-place, so fixing the shared function improves both. `writeFileDurable`'s cache callers are untouched.

Verified after: the victim file still reads `ORIGINAL SECRET CONTENT`, and the symlink itself has been replaced by a regular file. `TestWriteSecureFileDoesNotFollowSymlink` was proven red first:

```
Error: Not equal: expected: "ORIGINAL SECRET CONTENT" actual: "manifest: exported"
```

## Major - Export Template missing from the container menu

Fixed. Checked first whether exporting a container is even meaningful, since the alternative fix was deleting the dead branch: `resolveTemplateSource` already resolves `LevelContainers` to the parent Pod's manifest, so the container menu was simply missing the entry that dispatches to working code. Tests extended to cover the container menu in both the presence assertion and the hotkey-collision check.

## The four Minors

- `writeFileDurable`'s doc comment now covers both the kubectl cache directory and `<ConfigDir>/templates/`, and the stale "0644 on the file is fine" comment in `security_availability_cache.go` now says 0600.
- `userTemplateDir()` returns a distinct `errTemplateDirUnavailable` when `paths.ConfigDir()` fails, so an environment problem no longer surfaces as `invalid template name` and blames the resource.
- `docs/config-reference.md:614` now reads "no template-specific wrapper (keep the object's own `metadata`)", which no longer contradicts the example below it.
- `docs/keybindings.md` documents the destination picker's `j`/`k`, `c`/`f`/`t`, `Enter` and `Esc`/`q`.

## Verification

janosmiko:
The chip key was unreachable. Fixed in `ae485a6d`, along with the reason nobody noticed.

## `x` could never have worked

`internal/app/update_overlays.go:15` checks `isOverlayToggleKey` **before** the per-overlay handler. For `overlayAction` that returns true when the key equals `kb.ActionMenu`, which is `x`. So with the action menu open, `x` closed it and never reached the chip matcher at `update_overlays_misc.go:349`.

Pressing `x` twice opened the menu and immediately shut it. `x` is the one key that can never work as an action-menu chip.

## Why four passes missed it

The collision test compared chips **against each other**. Nothing compared a chip against the toggle key of the overlay hosting it - the wrong axis entirely, so the test passed, two reviewers agreed and this review had nothing to say.

The guard test is the real fix here. Reverting the chip to `x` now produces:

```
--- FAIL: TestActionMenuChipKeys_DoNotCollideWithHostToggleKey
    Error:    Should not be: "x"
    Messages: chip "Export Template" uses key "x", which is the action-menu toggle key:
              isOverlayToggleKey closes the overlay before the chip matcher runs
```

It asserts against `kb.ActionMenu` rather than a literal, so rebinding either side fails. A sibling test covers the keys `handleActionOverlayKey` consumes before chip matching - `esc`, `q`, `enter`, `up`/`k`/`ctrl+p`, `down`/`j`/`ctrl+n`, `ctrl+c`.

## The new keys

The user chose `T` for Export Template, which meant migrating what already held it:

| Action | Was | Now |
|---|---|---|
| Export Template | `x` | `T` |
| Log Top | `T` | `t` |
| Terminate Sync | `T` | `t` |
| Terminate Workflow | `T` | `t` |
| CronJob Trigger | `t` | `r` |

The last row was forced: CronJob already had `Trigger` on `t`, so moving `Log Top` there would have created a duplicate - the same class of bug. `Trigger` went to `r`, which across this file already means "make it run or run again": Restart, Resume Workflow, Reconcile, Force Refresh, Force Renew, Resize.

Node keeps `Taints` on `t`; it has no `Log Top`, so nothing collides. Verified across every menu builder: zero duplicate chip keys anywhere, and `kb.LogTop` - the separate log-viewer binding, still `T` - is untouched, since it lives in a different context.

## The silent failure that hid it

`exportTemplateCmd` returned a bare `nil` on two paths, so a wrong key produced no feedback at all. Both now report:

- `Export Template: security findings have no manifest to export`
- `Export Template: nothing selected to export`

A visible error would have made this obvious in seconds rather than surviving four reviews.

Docs updated together per the repo rule: `README.md`, `docs/keybindings.md`, the in-app help, and the test fixtures naming the old keys.

Gate: `gofmt`, `go build`, `go vet`, `golangci-lint` 0 issues, `go test -count=1 -race ./...` exit 0 across 26 packages.

janosmiko:
Follow-up in `c94d0e63`, marked `feat!` with a `BREAKING CHANGE` footer since three existing action keys move.

## Log Top is gone from Job and CronJob

Investigating the key collision turned up something worse than a key clash: **Log Top on a CronJob could never have worked.**

`kubectlGetPodSelector` (`internal/app/commands_logs.go`) has an explicit early return:

```go
// For CronJobs there's no direct pod selector.
if kind == "CronJob" {
    return ""
}
```

That empty selector forces the fallback, `kubectl logs cronjob/<name>`, which kubectl refuses. Verified against a live CronJob:

```
error: cannot get the logs from *v1.CronJob: selector for *v1.CronJob not implemented
```

Someone knew a CronJob has no selector, wrote the guard, and returned into a fallback that cannot work either.

Batch output also does not carry the HTTP access-log shape Log Top parses - it aggregates by method, host, path and status. So it is doubly wrong on both batch kinds, and it is now removed from Job and CronJob and from the `isGroupResource` gate in `update_actions_logtop.go`.

It remains on Pod, Service, Deployment, StatefulSet and DaemonSet - the long-running kinds that actually serve HTTP.

## CronJob keeps `t` for Trigger

Because Log Top left that menu, the earlier move of `Trigger` to `r` was no longer needed and has been reverted. `Trigger` is back on `t`, where muscle memory expects it.

## Final key map

| Action | Was | Now |
|---|---|---|
| Export Template | `x` (unreachable) | `T` |
| Log Top | `T` | `t` |
| Terminate Sync | `T` | `t` |
| Terminate Workflow | `T` | `t` |
| CronJob Trigger | `t` | `t`, unchanged |

Verified across every menu builder: zero duplicate chip keys.

## Not fixed here: TASK-892

`Tail Logs` and `Logs` on a CronJob fail for the same reason and are **left in place**. Fixing them means resolving a CronJob through the Jobs it owns to their pods, and choosing whether the view shows the latest run or every run - a behaviour change that belongs in its own PR rather than an export-template one. Filed as TASK-892 with the evidence.

Jobs are unaffected throughout: a Job does carry `spec.selector` with `batch.kubernetes.io/controller-uid`, so its log path already resolves.

Gate: `gofmt`, `go build`, `go vet`, `golangci-lint` 0 issues, `go test -count=1 -race ./...` exit 0 across 26 packages.

janosmiko:
On the two nitpicks from the latest review.

## Taking the `stripANSI(view)` one

Right, and it matters more than "trivial" suggests. Those tests assert on model fields and never prove the message reaches the screen - and the bug this branch just fixed was an **invisible failure**, where an unreachable chip produced no feedback at all. A test that only inspects `m.statusMessage` would not have caught it either. Being fixed with `basePush80Model()` and `stripANSI(view)`, per the repo convention.

## Not taking the consumed-keys one, with evidence

The proposal adds `g`, `G`, `ctrl+d`, `ctrl+u`, `ctrl+f`, `ctrl+b`, `pgup` and `pgdn` to the handler-consumed list. Applying it would fail the build on working code.

`handleActionOverlayKey` (`internal/app/update_overlays_misc.go`) consumes exactly:

```
esc, q, enter, up/k/ctrl+p, down/j/ctrl+n, ctrl+c
```

Everything else falls through to the chip matcher in `default`. None of the proposed keys are implemented in this overlay.

And `g` is **in use by two chips today**:

```
chips on 'g': ['Go to Node', 'Go to Pod']
```

Both work. Adding `g` to a list named "handler consumed keys" would assert they are unreachable, which is false, and turn the test red on correct behaviour.

The coding guideline being cited is real, but it describes the explorer and the text viewers. This overlay does not implement vim movement.

There is a genuine risk underneath, and it is the same class as the bug this branch fixes: if vim movement is ever added to this overlay, a chip on `g` breaks silently. The honest guard for that is a comment at the consumed-key list saying it must grow if the handler does - not an assertion that is untrue today. Adding that instead.

janosmiko:
The strip is now driven by field categories, with an overlay to choose them. Three commits: `d1bf213a`, `6c67941d`, `1cf14657`.

## What the strip was keeping

Probed against real cluster objects, not imagined. All of these survived and should not have:

| Kept before | Why it is wrong |
|---|---|
| `metadata.finalizers` | A template carrying one creates an object that cannot be deleted until a controller removes it - and never, if that controller is absent in the target cluster |
| `pod-template-hash` | Lands inside `spec.template.metadata.labels`, so the new Deployment starts with a stale hash its controller then fights. Same class as the Job `spec.selector` problem fixed earlier on this branch |
| `meta.helm.sh/*`, `heritage`, `release`, `chart`, `helm.sh/chart`, `app.kubernetes.io/managed-by: Helm` | The object claims membership of a Helm release it is not in. It may be adopted on the next `helm upgrade`, or deleted by `helm uninstall`. Most things worth exporting are Helm-managed, so this is the common case |
| `cni.projectcalico.org/podIP`, `field.cattle.io/publicEndpoints` | Live runtime status written into annotations - the Cattle one carries the actual hostname and addresses |

## Deny-lists, never prefix sweeps

`app.kubernetes.io/managed-by` is part of the standard recommended label set, so it is stripped **only when its value is `Helm`** - `managed-by: kustomize` is left alone, and there is a test for that.

The sharper case: a blanket `kubernetes.io/` sweep would eat `kubernetes.io/ingress.class`, which is author-written. Verified on a real object that it survives:

```yaml
metadata:
  annotations:
    kubernetes.io/ingress.class: nginx   # kept
    my-team/owner: platform              # kept
  labels:
    app: web                             # kept
    app.kubernetes.io/name: web          # kept
```

## The overlay

`s` from the destination picker opens a category picker. Six choosable rows - namespace, labels, annotations, Helm ownership, vendor runtime annotations, Secret values - with the correctness fields shown **locked** beneath an "always removed" divider, so you can see what goes without being offered a choice that yields a broken manifest.

Choices persist to `<state dir>/export_strip_prefs.yaml`. An absent key falls back to its own default rather than to "keep everything", so a partial or hand-edited file cannot silently disable the safety strips.

**The default path is unchanged at two keystrokes.** `TestExportStripOverlay_DefaultPathStaysTwoKeystrokes` asserts the manifest equals `k8s.StripToTemplate(doc)` byte for byte when the overlay is never opened. The new `s` key is also covered by the TASK-891 guard, so it cannot collide with the picker's own toggle key.

## Not stripped, deliberately

A real Service came through with `internalTrafficPolicy: Cluster`, `sessionAffinity: None`, `type: ClusterIP`. They are API defaults and harmless, but indistinguishable from a deliberate choice - stripping them would delete intent, which is the failure mode this feature guards against everywhere else.

## Tests

Every category asserts **both halves**: the key is gone AND an author-written sibling survived. The second half is what catches an over-broad strip.

Each was proven red first, for example the Helm case failing seven times, once per marker. The `stripANSI(view)` additions were checked for vacuity independently of the model-field assertions.

## File caps

`styles.go` and `app.go` are untouched at 800. The overlay state lives inside `exportTemplateState` rather than on `Model`. `view_overlays.go` did reach 804, so the two export render branches were collapsed into one helper in the new file; it is back to 799.

Gate: `gofmt`, `go build`, `go vet`, `golangci-lint` 0 issues, `go test -count=1 -race ./...` exit 0 across 26 packages.

janosmiko:
Taken, in `c2f2f784` - but with the repo's own pairing rather than the suggested diff.

The overlay had `j`/`k`, arrows and `g`/`G`, so `ctrl+d`, `ctrl+u`, `ctrl+f`, `ctrl+b`, `pgup` and `pgdown` did nothing here while every other overlay moved. Six rows do not need half-page scrolling today, but an inconsistent cursor is exactly what breaks muscle memory, and "the list is short" is the reasoning that ages badly if categories grow.

## One change to the proposal

The suggested diff pairs `pgdown` with `ctrl+d` (half page). The rest of the repo pairs them the other way:

```go
case "ctrl+d", "shift+down":   // half page
case "ctrl+f", "pgdown":       // full page
```

Applying it verbatim would have made this the one overlay that disagrees with the others - the opposite of what the finding is for. Followed the existing convention instead, so `shift+down` is the half-page alias and `pgdown` the full-page one.

Also confirmed `pgdown` is this repo's spelling: 62 uses, no `pgdn`.

## Test

`TestExportStripKey_SupportsFullVimNavigation`, proven red before the fix:

```
--- FAIL: TestExportStripKey_SupportsFullVimNavigation/ctrl+u_half_page_up
    Messages: ctrl+u should move the cursor to row 2
--- FAIL: TestExportStripKey_SupportsFullVimNavigation/ctrl+f_full_page_down
    Messages: ctrl+f should move the cursor to row 5
```

Seven cases, each asserting the cursor lands on a specific row rather than merely moving - a test that only checked "the cursor changed" would pass on a single-step fallback.

Hint bar gained `ctrl+d/u: half page`.

Gate: `gofmt`, `go build`, `go vet`, `golangci-lint` 0 issues, `go test -count=1 -race ./...` exit 0 across 26 packages.

PR #619, opened 2026-08-12
feat: delete saved user templates from the template picker

janosmiko:
The Create-from-Template picker listed templates from `~/.config/lfk/templates/` but offered no way to remove one, so a template saved by mistake could only be cleared with `rm`.

`d` on a `User` row now confirms through the same `overlayConfirm` box every destructive action uses, removes the backing file, and refreshes the list in place. Cancelling returns to the picker with the file untouched. A built-in row refuses with "Only your own templates can be deleted", because there is no file behind it. Both `.yaml` and `.yml` are tried, since the picker lists both.

Two decisions worth naming:

- **No new Model field.** `app.go` sits at its 800-line cap and the field pushed it to 802. The template name rides in `confirmAction`, which already means "the thing being acted on" and is used that way by every other confirm opener.
- **The confirm skips the read-only net and the loading spinner.** A template file is local configuration, not a cluster mutation, so `"Delete Template"` is deliberately absent from `mutatingActions`.

A failed `os.Remove` now outranks a partial success: reporting the file that came off while staying quiet about the one left behind would put the picker and the directory out of step with no sign of it. That ordering has no test - removal permission lives on the directory, not the file, so there is no portable way to make one removal fail and the other succeed. Called out rather than papered over.

## Test plan

Every test was proven red before the fix, per test by name.

- [x] `TestDeleteUserTemplate_RemovesFile`, `_RemovesYmlFile`, `_RejectsPathTraversal`, `_ReportsMissingTemplate`
- [x] `TestTemplateOverlayKey_DConfirmsUserTemplate` - red: `"" does not contain "web"`
- [x] `TestTemplateOverlayKey_DRefusesBuiltinTemplate` - red: `"" does not contain "own templates"`
- [x] `TestConfirmOverlayKey_CancelledTemplateDeleteReturnsToPicker` - red: overlay not equal
- [x] `TestConfirmOverlayKey_TemplateDeleteRemovesFileAndRefreshesPicker` - red: `web.yaml exists`, `Should not be: "User"`
- [x] `go build ./...`, `go test ./internal/app/...`, `golangci-lint run ./internal/...` - 0 issues
- [ ] Manual: `a`, then `d` on a saved template; confirm, cancel, and try `d` on a built-in

**Cursor-clamp test written and deleted.** `mergedTemplates` always appends the built-ins, so the list never shrinks far enough for the clamp to bite - the test passed with and without the fix. The clamp call stays because it is correct and free, but it carries no honest guard.

Closes TASK-894.


janosmiko:
Good catch, and it was worse than a nitpick - confirmed with a probe before fixing:

```
row name="web" desc="Deployment"     <- web.yaml
row name="web" desc="Service"        <- web.yml
rows after delete: 0                 <- pressed d once
```

Two files with one base name are two separate templates, and the picker lists them as two rows. Deleting by name took both, so one confirmation destroyed a row the user never selected. That is silent data loss, not an edge case.

Fixed the way you suggested: `ResourceTemplate` now carries the file it was read from, and the confirmation holds that path instead of the name. `deleteUserTemplate` re-checks the directory before removing, since the path reaches it through Model state.

`TestTemplateDelete_SameBaseNameKeepsTheOtherFile` proven red against the old delete logic - keeping the new `Path` field so the failure is behavioural rather than a compile error:

```
--- FAIL: TestTemplateDelete_SameBaseNameKeepsTheOtherFile
    Error:    unable to find file ".../templates/web.yml"
    Messages: the unselected row survives
```

Gate green: build, `internal/app` + `internal/model`, `golangci-lint run ./internal/...` 0 issues. Pushed as c29af7a8.

Still open and unrelated to this fix: the picker shows both rows as plain `web`, so they are told apart only by the description column. That naming problem is TASK-890.

PR #620, opened 2026-08-12
fix: resolve CronJob logs through its newest Job's pods

janosmiko:
Tail Logs and Logs on a CronJob have never worked. `kubectlGetPodSelector` returned an empty selector for kind CronJob by design, which forced the fallback `kubectl logs cronjob/<name>`, and kubectl rejects that outright:

```
error: cannot get the logs from *v1.CronJob: selector for *v1.CronJob not implemented
```

Verified against a live CronJob before the fix: `spec.selector` is absent and the fallback errors exactly as above. Both log actions offered something the code could not deliver.

A CronJob now resolves to the pods of its newest owned Job. **Most recent Job only** - "logs" on a scheduled task means the last run, and aggregating every historical Job would mix runs with no marker between them. Jobs are untouched, they already carry `spec.selector` and resolve through the existing path.

Older clusters use unprefixed `job-name` / `controller-uid`, so the prefixed pair is tried first and the legacy pair second. A CronJob that has never run, or whose Jobs were cleaned up, reports "no runs yet" through the status bar rather than opening an empty viewer or leaking a raw kubectl error.

**Ownership is matched by UID, not name.** Review caught the first version matching `ownerReferences[].name`, which silently attributes an orphaned Job's logs to a same-named replacement: delete a CronJob with `--cascade=orphan`, create a new one under the same name, open its logs before its first run, and you would read the deleted CronJob's output under the new resource's name. The UID comes from a live `kubectl get cronjob`; a failed read returns "no runs yet" rather than falling back to names, and an empty UID is rejected outright.

`startMultiLogStream` / `restartMultiLogStream` moved to `commands_logs_multi.go` unchanged, to bring `commands_logs.go` under the 800-line cap. Diffed byte-for-byte against the parent commit.

Log Top stays out: it was removed from Job and CronJob in the change that found this, because batch output has none of the HTTP access-log shape it parses.

## Test plan

Every test proven red before the fix, per test by name.

- [x] `TestStartLogStream_CronJob_UsesNewestJobSelector` - red with the reported bug reproduced verbatim: argv was `logs cronjob/my-cron --all-containers=true`, no `-l`
- [x] `TestLatestOwnedJob_IgnoresJobOwnedByDifferentCronJob` - red: `found=false`
- [x] `TestLatestOwnedJob_OrphanedJobSameNameDifferentUID_NotMatched` - red: old code matched the orphan by name
- [x] `TestLatestOwnedJob_TiebreakOnEqualTimestamp` - red: `want my-cron-200, got my-cron-100`
- [ ] Manual, and the gap that matters: **no live cluster.** The selector logic and the legacy-label fallback were checked against fake `kubectl` scripts, not a real CronJob to Job to Pod chain

`TestStartLogStream_Job_StillUsesKubectlGetPodSelector` passes on unfixed code too. That is correct - the Job path was never broken, so it guards against future breakage rather than proving this fix.

Known and out of scope: multi-select log streaming still has no CronJob branch and would build the same rejected `logs cronjob/<name>`.

Closes TASK-892.

### DIFF
--- a/internal/app/commands.go
+++ b/internal/app/commands.go
@@ -219,10 +219,18 @@ func (m Model) loadPodsForLogAction() tea.Cmd {
 	kubeconfigPaths := m.client.KubeconfigPathForContext(kctx)
 
 	return func() tea.Msg {
-		// Get the selector for this parent resource.
-		selector := kubectlGetPodSelector(kubectlPath, kubeconfigPaths, ns, kind, name, m.kubectlContext(kctx))
-		if selector == "" {
-			return podLogSelectMsg{err: fmt.Errorf("could not determine pod selector for %s/%s", kind, name)}
+		var selector string
+		if kind == "CronJob" {
+			resolved, ok := resolveCronJobPodSelector(kubectlPath, kubeconfigPaths, ns, name, m.kubectlContext(kctx))
+			if !ok {
+				return podLogSelectMsg{err: errCronJobNoRuns}
+			}
+			selector = resolved
+		} else {
+			selector = kubectlGetPodSelector(kubectlPath, kubeconfigPaths, ns, kind, name, m.kubectlContext(kctx))
+			if selector == "" {
+				return podLogSelectMsg{err: fmt.Errorf("could not determine pod selector for %s/%s", kind, name)}
+			}
 		}
 
 		// Fetch pods matching the selector.
--- a/internal/app/commands.go
+++ b/internal/app/commands.go
@@ -221,9 +221,9 @@ func (m Model) loadPodsForLogAction() tea.Cmd {
 	return func() tea.Msg {
 		var selector string
 		if kind == "CronJob" {
-			resolved, ok := resolveCronJobPodSelector(kubectlPath, kubeconfigPaths, ns, name, m.kubectlContext(kctx))
-			if !ok {
-				return podLogSelectMsg{err: errCronJobNoRuns}
+			resolved, err := resolveCronJobPodSelector(kubectlPath, kubeconfigPaths, ns, name, m.kubectlContext(kctx))
+			if err != nil {
+				return podLogSelectMsg{err: err}
 			}
 			selector = resolved
 		} else {
--- a/internal/app/commands_logs.go
+++ b/internal/app/commands_logs.go
@@ -135,10 +135,10 @@ func (m *Model) startLogStream() tea.Cmd {
 		}
 		switch kind {
 		case "CronJob":
-			selector, ok := resolveCronJobPodSelector(kubectlPath, kubeconfigPaths, ns, name, m.kubectlContext(kctx))
-			if !ok {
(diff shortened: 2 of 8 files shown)


janosmiko:
Both Majors were real. Fixed in b8aea0c6.

**Every kubectl failure read as "no runs yet".** The resolver collapsed an RBAC denial, expired credentials and an apiserver outage into the same answer as a CronJob that has never run. The resolver now returns `error` rather than `bool`, `errCronJobNoRuns` is reserved for a genuinely empty result, and everything else reaches the status bar with the underlying stderr folded in.

The red proof needed isolating, and it was worth doing. The whole test file fails to compile against the old `bool` API, which proves only that the signature changed. Moving the caller-level test into a file of its own and reverting the four production files gives the behavioural failure:

```
--- FAIL: TestStartLogStream_CronJob_KubectlFailure_SurfacesRealError
    Error:    Should not be: "\x00lfk:cronjob-no-runs\x00my-cron"
              a real kubectl failure must not read as the no-runs marker
    Error:    "No runs yet for CronJob my-cron" should not contain "No runs yet"
    Error:    "No runs yet for CronJob my-cron" does not contain "Forbidden"
```

That is the bug in one line: `Forbidden` in, `No runs yet` out.

**The no-runs marker never fired in a background tab.** The check sat after the branch that returns for a background channel, so that tab appended the raw NUL-wrapped sentinel to its buffer as a visible log line. Now handled inside the branch, still draining `msg.ch`. Red: `Should be empty, but was [\x00lfk:cronjob-no-runs\x00my-cron]`.

Gate re-verified by me in the worktree: build clean, `go test ./...` all packages, `golangci-lint run ./...` 0 issues.

Not fixed here, on purpose: the Minor on `commands_logs_multi.go`. That function was moved verbatim in this PR and the missing status message predates it, so fixing it here would bury a pre-existing bug inside a CronJob change. Filed as TASK-895.

janosmiko:
Agreed on the gap, not on fixing it here. The PR body already names it: the selector logic and the legacy-label fallback were checked against fake `kubectl` scripts, not a real CronJob to Job to Pod chain.

The reason it stays out of this PR is that the E2E lane you are asking for does not exist yet. `e2e/demo_test.go` drives the built binary in a pty under `--demo`, so it talks to the demo fake and never to an apiserver, and no workflow in `.github/workflows/` runs it - `make e2e` is manual. "Add a repeatable CI E2E scenario" therefore means standing up a throwaway cluster in CI, adding an E2E job, and then writing the scenario. That is test infrastructure, not a CronJob fix, and burying it in this branch would make both harder to review.

Filed as TASK-896, in two steps so the cheap half lands first:
1. Give the demo cluster a CronJob that owns Jobs and Pods, and add an e2e scenario that opens its logs. Catches wiring breakage, still not real kubectl.
2. Add a CI job that starts kind or k3d and runs the same scenario against it, which is what actually validates the label semantics and the legacy fallback.

Every kubectl-shaped path in this repo shares that gap, so scoping the fix to the CronJob change would have left the other paths uncovered while looking like the problem was solved.

PR #621, opened 2026-08-12
fix: key saved templates by namespace, not just name

janosmiko:
> Builds on the `Path` field #619 added to `ResourceTemplate`. #619 is merged, so this targets `main` and carries only its own three commits.

`saveUserTemplate` wrote `<ConfigDir>/templates/<metadata.name>.yaml`. Two objects with the same name in different namespaces or clusters wrote the same file, and the second overwrote the first with no warning. A Deployment called `nginx` in staging and one in prod is a plausible pair.

The file name now carries the namespace: `<namespace>__<name>.yaml`, or `_cluster__<name>.yaml` for a cluster-scoped kind. A **double underscore** separates them because `_` is illegal in a DNS-1123 object name and namespace, so the split back is unambiguous for any file lfk wrote. The picker still shows the object name and gains the namespace beside the Kind in its description column.

Two decisions worth stating, because neither is the obvious one:

- **The cluster context is deliberately not in the file name.** EKS contexts are ARNs containing `:` and `/`. Encoding those into a path has no clean escape.
- **So the residual collision gets a confirmation instead.** The same namespace and name in two different clusters now asks before replacing the file. Qualified naming removes the common case, the prompt covers the rest.

Declining that prompt returns to the destination picker with the export still staged, so saying no to an overwrite leaves clipboard and file reachable without exporting again. The delete confirm already returned to its opener; this now matches.

**Backward compatible.** A hand-authored file with no `__` keeps its whole base name, which matters because users keep these in dotfiles. `deleteUserTemplate` needed no change - it deletes by `Path`.

## Test plan

Every test proven red before its fix, per test by name.

- [x] `TestSaveUserTemplate_DifferentNamespacesSameNameBothSurvive` - red: 1 of 2 files existed
- [x] `TestSaveUserTemplate_ClusterScoped` - red: `worker-1.yaml` not found
- [x] `TestApplyExportTemplatePicker_ExistingFileOpensConfirm`, `TestConfirmOverlayKey_AcceptedOverwriteReplacesFile` - red: no confirm, no cmd, no save
- [x] `TestConfirmOverlayKey_CancelledOverwriteKeepsFile` - red on the return path: `declining the overwrite returns to the destinations`, `the export stays staged`
- [x] `TestLoadUserTemplates_TrailingSeparatorKeepsWholeBaseName` - red: `"[]" should have 1 item(s), but has 0`
- [ ] Manual: save two same-named objects from different namespaces, then save one twice to hit the confirm

**Known limit, documented in the code rather than fixed:** the `__` split also applies to hand-authored names, which are under no DNS-1123 rule. A file called `my__template.yaml` reads as namespace `my`. It affects only the two columns the picker draws - no data loss, no path escape. A file ending in the separator (`web__.yaml`) used to vanish from the picker entirely, because an empty name is dropped; it now falls back to the whole base name.

Review confirmed the path composition is safe: namespace and name are sanitized and validated **separately** before joining, so neither half can carry a separator into the result, and the overwrite check and the write share one function so they cannot disagree about the target.

Closes TASK-890.

### DIFF
--- a/internal/app/commands_export_template.go
+++ b/internal/app/commands_export_template.go
@@ -24,7 +24,7 @@ func (m Model) exportTemplateCmd() (Model, tea.Cmd) {
 		m.setStatusMessage("Export Template: security findings have no manifest to export", true)
 		return m, scheduleStatusClear()
 	}
-	fetch, name, kind, ok := m.resolveTemplateSource()
+	fetch, name, kind, namespace, ok := m.resolveTemplateSource()
 	if !ok {
 		m.setStatusMessage("Export Template: nothing selected to export", true)
 		return m, scheduleStatusClear()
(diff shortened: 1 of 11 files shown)


janosmiko:
Both taken.

**Major, and correct: the write ran inside key handling.** `saveUserTemplate` fsyncs the file and its directory, so a slow filesystem stalled the whole UI after the confirmation. It now returns a `tea.Cmd` emitting `actionResultMsg`, matching the non-overwrite path it should have matched from the start. Proven red by reverting to the synchronous version: `Should be true` on the assertion that a command comes back.

**Minor, also real:** `templateSavePath` returned a bare `false` for both a missing config directory and a rejected name, so a $HOME problem was reported as an invalid template name. It returns an `error` now, and the caller propagates whichever it gets. Red proof:

```
--- FAIL: TestTemplateSavePath_DistinguishesFailures/directory_unavailable
    Error: Target error should be in err chain:
```

That branch turned out to be reachable in a test after all - clearing `LFK_CONFIG_DIR`, `HOME` and `XDG_CONFIG_HOME` makes `userTemplateDir()` return empty - so it has a real guard rather than a note.

Gate green: build, `go test ./...`, `golangci-lint run ./...` 0 issues. Pushed as 50d59916.

PR #622, opened 2026-08-12
fix: report a status message when every multi-log stream fails to start

janosmiko:
> `commands_logs_multi.go` arrived in #620, which is merged, so this targets `main` and carries only its own commit.

`startMultiLogStream` skipped an item on a pipe or start failure with only a `logger` call. When every item failed, `wg.Wait()` returned at once, `close(ch)` fired, and the consumer received `logLineMsg{done: true}`. The user got an empty log viewer and no reason for it.

It now counts the streams that actually started. At zero it cancels the context, restores the view mode the model was in before the call, and reports through `setStatusMessage` with the item count and the last start error. A partial failure is untouched: the viewer still opens with the streams that did start.

No synchronisation was needed. The counter is written only in the calling goroutine, before any per-item goroutine is spawned, and read once after the loop - confirmed by review and by a clean `-race` run.

This predates #620, which only moved the function to `commands_logs_multi.go` unchanged. It was raised by CodeRabbit there and kept out to hold that PR's scope.

## Test plan

- [x] `TestStartMultiLogStream_AllFail_ReportsStatusAndKeepsMode` - proven red: mode stayed `modeLogs` instead of `modeExplorer`, `statusMessageErr` was false, `statusMessage` empty
- [x] `go build ./...`, `go test ./...` (all packages), `go test -race ./internal/app/` clean, `golangci-lint run ./...` - 0 issues
- [ ] Manual: select several pods, break `kubectl` on PATH, run the bulk Logs action

**One test, and that is the honest ceiling here.** Two more were written for the other acceptance criteria - partial failure still opens the viewer, all-success unchanged - and both passed against the unfixed baseline, so they were deleted rather than kept as decorative guards. Review reached the same conclusion independently: the `continue`-on-error loop body is unchanged by this commit, so no non-vacuous test of the partial path is constructible from this diff.

Closes TASK-895.

### DIFF
--- a/internal/app/commands_logs_multi.go
+++ b/internal/app/commands_logs_multi.go
@@ -25,6 +25,8 @@ func (m *Model) startMultiLogStream(items []model.Item) (tea.Model, tea.Cmd) {
 		return m, func() tea.Msg { return actionResultMsg{err: fmt.Errorf("kubectl not found: %w", err)} }
 	}
 
+	previousMode := m.mode
+
 	// Initialize log viewer state.
 	m.mode = modeLogs
 	m.resetLogBuffer()
@@ -55,6 +57,8 @@ func (m *Model) startMultiLogStream(items []model.Item) (tea.Model, tea.Cmd) {
 	ns := m.resolveNamespace()
 
 	var wg sync.WaitGroup
+	started := 0
+	var lastErr error
 	for _, item := range items {
 		itemCtx := kctx
 		if item.ClusterName != "" {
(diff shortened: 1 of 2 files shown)


janosmiko:
Taken - and it is the assertion class I have been cutting all session, so thank you for catching mine.

`Contains(msg, "3")` passes on any message carrying that digit anywhere, including a pod name or a timestamp. Now:

```go
strings.HasPrefix(mdl.statusMessage, "Failed to start logs for 3 resources:")
```

The trailing colon matters: it holds the appended cause in place, so a regression that drops the underlying start error fails the test rather than passing quietly.

Gate green: build, `go test ./...`, `-race` clean, `golangci-lint run ./...` 0 issues. Pushed as acc0241f.

PR #623, opened 2026-08-12
test: cover the CronJob log path in demo and on a real kind cluster

janosmiko:
> The scenario needs the CronJob log fix from #620, which is merged, so this targets `main` and carries only its own three commits.

The e2e suite drove the built binary in a pty under `--demo`, behind the `e2e` build tag. It talked to the demo fake, never to an apiserver, and no workflow ran it - `make e2e` was manual. That left every kubectl-shaped path checked only against fake `kubectl` scripts.

**It found a real bug on its first run.** The demo kubectl emulation had no handler for `get cronjob NAME`, `get jobs`, or `get pods -l SELECTOR`. So the CronJob log fix in #620, which passes against fake `kubectl` scripts, failed under `--demo` with `cronjob uid missing`. A unit test could not have caught this, because the fake scripts stand in for exactly the layer that was missing.

Two commits.

**`test:` demo coverage.** Seeds a CronJob to Job to Pod chain (`nightly-backup`), teaches the demo kubectl emulation the three calls it was missing, and adds `TestDemoModeCronJobLogs`. Reuses the existing pty driving rather than adding a second harness. This catches wiring breakage and nothing more - the test says so in its own comment.

**`ci:` a real cluster.** A new workflow provisions a kind cluster, applies a CronJob/Job/Pod fixture with server-assigned UIDs substituted in between calls, and runs `TestClusterModeCronJobLogs`. `ci.yml` is untouched, every action is pinned to a commit SHA, the job has a 15-minute bound, and teardown runs through the action's own `post` step regardless of outcome.

**The label coverage is split on purpose.** The demo pod carries only the modern `batch.kubernetes.io/` labels. The kind fixture carries only the legacy unprefixed `job-name` / `controller-uid` pair, so the CI lane is what finally forces the fallback branch of `resolveCronJobPodSelector` to be the one that matches. That branch has never been exercised against a real apiserver until now.

kind over k3d: no k3d tooling exists in this repo, and the thing under test is controller-manager label semantics, where fidelity to upstream matters.

## Test plan

- [x] `make e2e` - real run, `TestDemoModeStartup` PASS, `TestDemoModeCronJobLogs` PASS (3.67s), `TestClusterModeCronJobLogs` SKIP without `LFK_E2E_REAL_CLUSTER`
- [x] `actionlint` on the workflow, including shellcheck over the `run:` blocks - 0 issues
- [ ] **The CI job itself has never run.** It needs this PR to exercise it

**Unverified, and it cannot be otherwise from a sandbox:** whether kind provisioning leaves `kubectl` and `lfk` pointed at the same kubeconfig on a real runner, whether the UID-substitution `kubectl apply` sequence behaves as scripted against a live apiserver, whether the pty navigation timing holds against slower RBAC-checked startup, and whether the 30-second `waitFor` budget is enough on `ubuntu-latest`. Watch this PR's own check.

One existing test changed: `TestTicker_FullCycleInvariants` counted Running pods cluster-wide, and the new static Pod tipped its majority. Narrowed to `NamespaceDemo`, which is the only namespace the ticker mutates - a fixed pod elsewhere dilutes a ratio it cannot move. Both invariants the test asserts are intact, and the comment describing them was restored after the narrowing dropped one.

Closes TASK-896.

### DIFF
--- a/.github/workflows/e2e-cluster.yml
+++ b/.github/workflows/e2e-cluster.yml
@@ -25,6 +25,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          # Nothing after this step talks to git, so the token has no reason to
+          # stay readable in the local git config.
+          persist-credentials: false
 
       - name: Set up Go
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
--- a/.github/workflows/e2e-cluster.yml
+++ b/.github/workflows/e2e-cluster.yml
@@ -1,0 +1,144 @@
+name: E2E (real cluster)
+
+# Runs the e2e suite against a real kind apiserver instead of the --demo
+# fake, so the CronJob log label selectors validate for real (TASK-896).
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  e2e-cluster:
+    name: e2e against a kind cluster
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    env:
+      LFK_E2E_REAL_CLUSTER: "1"
+      LFK_E2E_LOG_MARKER: lfk-e2e-cronjob-log-marker
+      LFK_NAMESPACE: lfk-e2e
+      LFK_CRONJOB: nightly-backup
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Set up Go
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+        with:
+          go-version-file: go.mod
+
+      # Cleanup (kind delete cluster) is this action's own post step, which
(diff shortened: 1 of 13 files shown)


janosmiko:
Both taken, plus a third thing I found while verifying them.

**`persist-credentials: false`** - correct, nothing after checkout talks to git.

**The UID assertion** - real, and the same weak-assertion shape as the finding on #622. Rejecting only the seeded UID passed on any other non-empty value, and a non-empty UID is exactly what makes an unknown CronJob look real to `getCronJobUID`. It now decodes the JSON and requires an empty `metadata.uid`. Proven red by making the emulation return `some-other-uid`:

```
metadata.uid = "some-other-uid", want empty for an unseeded CronJob name
```

**The third thing: this suite was load-flaky, and I nearly shipped it.** While gating three worktrees at once, `make e2e` failed at 26.2s against a ~10.5s norm. Eleven runs on an idle machine then passed, which is how a flake like this survives review. A pty `waitFor` is wall-clock, so a contended machine eats the budget while the app is fine - and a shared CI runner is no less contended than a laptop running three test suites. `startupTimeout` 15s to 45s, `clusterStartupTimeout` 30s to 90s. Re-ran under deliberate load (three concurrent `go test ./...`) and it passes at 9.3s.

That was the risk I listed as unverified when this PR went up. It reproduced locally before CI ever saw it.

Gate green: build, `go test ./...`, `golangci-lint run ./...` 0 issues, `actionlint` clean, `make e2e` passing. Pushed as 13055b1b.

The CI job itself is still unrun. Its first execution is this PR's own check.

PR #631, opened 2026-08-17
feat: add a Changed column with the time since the last change

janosmiko:
## Summary

- Every resource list gains a `CHANGED` column: the time since the object last changed, taken from the newest condition transition, the container restart time, or the newest `managedFields` write. Those fields already arrive in the list response, so the column costs no extra API call and fills on the first render.
- A row with none of those timestamps renders an empty cell and sorts last in both directions, so kinds that cannot answer the question stay in name order.
- The column is pinned ahead of the metrics block. Behind it, a narrow pane spent the width budget on CPU and MEM and the column vanished from the list the user had just sorted by it.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run` on `internal/...` and `e2e/...`, 0 issues
- [x] `go test ./...`
- [x] `go test ./... -race`, 26 packages
- [x] `make e2e`, including a new pty-driven check that the column renders and fills for several rows
- [x] `BenchmarkApplyChangedColumn`: 1.24 ms per 5000 rows, no API calls
- [ ] Manual: `./lfk --demo`, press `2`, every pod shows a `CHANGED` value
- [ ] Manual: a real cluster. The demo fake cluster never exercises the `managedFields` parse on live data.

### DIFF
--- a/e2e/demo_test.go
+++ b/e2e/demo_test.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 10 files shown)


janosmiko:
Addressed two of three. Declining the third, reasoning below.

**1. e2e counted redraw frames, not pods — fixed.** Confirmed and it was the serious one. `out.String()` holds every repaint, so `filled++` per matching line let one populated pod clear a three-row threshold on its own. The check would have passed in exactly the state it exists to catch: the column filled for only the pod whose restart lfk happened to witness. It now deduplicates by pod name and reports the names it found on failure. Still passes, so the behaviour was right and only the assertion was weak.

**2. `ActiveSortableColumnCount` not restored — fixed.** `RenderTable` writes it, the test restored only `ActiveSortableColumns`. Applied as suggested.

**3. Return a copy instead of enriching in place — declining.**

- Every peer enrichment in this package mutates the caller's slice: `applyViewColumns`, `carryOverMetricsColumnsFrom`, `carryOverServiceEndpointColumns`, and `sortItemsByColumn` itself. Returning a new slice from this one function alone would make it the odd one out, and `sortMiddleItems` / `sortPreviewItems` both exist to work on a slice their caller owns.
- The immutability guideline is a common-layer default that the Go-specific rules explicitly override where in-place mutation is idiomatic.
- The real aliasing hazard here is narrower and already handled: tab snapshots shallow-copy `Item`, so the function must never write *through* the existing `Columns` slice. It builds a fresh slice and assigns it, which is the invariant that matters.

Happy to revisit if you would rather change the whole family of enrichment functions together, but not this one in isolation.

PR #641, opened 2026-08-18
fix: sort the Changed and Uptime columns by their real timestamps

janosmiko:
## Summary

- The `Changed` column sorted on its rendered cell. `FormatAge` buckets a range of ages into one string, so a pod changed 60s ago and a pod changed 61s ago both read `1m` and the order flipped to name order. `comparePrimaryColumn` now compares the `changeAt()` timestamp.
- The `Uptime` column had the same defect one bucket wider: `5d` covers a full day. `model.Item` now carries `BootedAt`, stamped by the Prometheus enrichment and carried across watch ticks, and the comparator reads it.
- Rows with no timestamp keep the string comparison, so a CRD `additionalPrinterColumns` field named `Uptime` is unaffected. `n/a` rows still sort last in both directions.

`Age` and the events `Last Seen` column already compared item timestamps. `Duration` keeps second precision, so it has no bucket collision.

## Test plan

- [x] `TestChangedSortUsesTheRealTimeNotTheRenderedCell` - two pods at 60s and 61s, both cells `1m`
- [x] `TestUptimeSortUsesTheBootTimeNotTheRenderedCell` - two nodes inside the `5d` bucket
- [x] `TestUptimeSortFallsBackToTheCellWithoutABootTime` - CRD column regression guard
- [x] `TestUptimeEnrichmentStampsTheBootTime`, `TestUptimeCarryOverKeepsTheBootTime`
- [x] `go build ./...`, `go test ./...`, `golangci-lint run ./internal/...` all clean

### DIFF
--- a/internal/app/changed_column_test.go
+++ b/internal/app/changed_column_test.go
@@ -285,3 +285,23 @@ func TestChangedJoinsTheSortCycle(t *testing.T) {
 		t.Fatalf("the Changed key must be reachable in the cycle, got idx=%d ok=%v", idx, ok)
 	}
 }
+
+func TestChangedSortUsesTheRealTimeNotTheRenderedCell(t *testing.T) {
+	now := time.Now()
+	// Both rows render "1m". Only the timestamp separates them.
+	items := []model.Item{
+		podWithConditions("aaa-older", now.Add(-61*time.Second)),
+		podWithConditions("zzz-newer", now.Add(-60*time.Second)),
+	}
+
+	applyChangedColumn(items, now)
+	if a, b := items[0].ColumnValue(ChangedColumnKey), items[1].ColumnValue(ChangedColumnKey); a != b {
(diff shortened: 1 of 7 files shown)


janosmiko:
Fixed the per-item clock read in fe79ea6 with a regression test (`TestUptimeEnrichmentStampsOneReferenceTime`).

Skipping the immutability finding. `updateNodeUptimeEnriched` already mutates `m.middleItems` in place through `item := &m.middleItems[i]`, and every sibling enrichment handler in that file does the same. `BootedAt` is a scalar on an Item value, so `cloneCurrentTab` snapshots do not share it - the shared-backing-array hazard applies to the `Columns` slice, which this change replaces rather than writes through. Reshaping the enrichment path is a separate refactor.

janosmiko:
@coderabbitai review

PR #643, opened 2026-08-19
fix: sort relative-time columns by duration

janosmiko:
## Summary

Sorting by `Last Transition` put `1138d ago` above `14d ago`. The cells carry `formatRelativeTime` output, no comparator matched that shape, and the fallback compared the strings, where `"11..." < "14..."`. An audit of every column found four more with the same defect.

- `compareRelativeAgoCmp` parses the duration out of a `"1138d ago"` cell. Registered for `Last Transition` (generic CRDs and Flux) and `Synced At` (Argo, whose cell can read `syncing 5m ago`).
- `Last Scale Time` (HPA), `Next` (CronJob) and `Last Deployed` (Helm) hold `formatAge` output and now use `compareUptimeCmp`, so `9h` sorts before `10d`.
- `Interval` (Flux) holds a Go duration string and now uses `compareDurationCmp`.
- `Last Seen` compares the event timestamp. Only the default Event sort reached that comparator before; picking the column by hand fell through to the rounded cell.
- A cell the sort cannot read now lands at the bottom in both directions, through the same path the `n/a` metrics rows use.

Ruled out and left alone: `Expires`, `Renewal` and `Last Schedule` hold raw RFC3339 with a `Z` offset, and that text sorts chronologically.

## Test plan

- [x] `TestLastTransitionSortsByDurationNotLexically` - the reported case, `1138d` against `14d`
- [x] `TestAgeShapedColumnsSortByDuration` - `10d` against `9h` across the four `formatAge` columns
- [x] `TestLastSeenColumnPrefersTheEventTimestamp` - two events one second apart sharing a `1m` cell
- [x] `TestIntervalSortsByDuration`, `TestSyncedAtSortsThroughItsStatusPrefix`, `TestLastTransitionOrdersEveryUnit`
- [x] Unreadable cells sort last in both directions: `TestLastTransitionKeepsUnparseableCellsLast`, `TestAgeShapedColumnsKeepUnreadableCellsLast`
- [x] `go build ./...`, `go test ./internal/...`, `golangci-lint run ./internal/...` clean

## Known limit

Past 24h a `formatAge` cell is day-grained, so two resources that changed on the same day still fall through to the name tiebreaker. Fixing that needs the raw timestamp on the Item, which is a larger change.

### DIFF
--- a/internal/app/relative_time_sort_test.go
+++ b/internal/app/relative_time_sort_test.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"testing"
+	"time"
 
 	"github.com/janosmiko/lfk/internal/model"
 )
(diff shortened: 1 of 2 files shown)


janosmiko:
Fixed in dc4252e. `metricValueMissing` now classifies an unparseable `Last Transition` or `Synced At` cell as missing, so it sorts last before the direction flip. `TestLastTransitionKeepsUnparseableCellsLast` covers `""`, `"unknown"` and `"n/a"` in both directions.

janosmiko:
@coderabbitai review

janosmiko:
Fixed in acbd35f. `readableTimeValue` maps each time-shaped column to its parser, and `metricValueMissing` uses it, so an unreadable cell in `Last Scale Time`, `Next`, `Last Deployed`, `Last Seen` or `Interval` sorts last before the direction flip. `Interval` joined `metricsMissingLastColumns`. `TestTimeShapedColumnsKeepUnreadableCellsLast` covers all seven columns against `""`, `"n/a"` and `"unknown"` in both directions - 42 cases.

PR #645, opened 2026-08-19
docs: restore the Star History chart

Dessalines39394:
The Star History chart was removed from the README in commit 92a4d83 ("docs: doc-audit fixes and README reshape", PR #630). This restores it in the same place and points it at a working endpoint, since the previous chart is currently broken because of GitHub stargazer API restrictions. The fix is necessary so the project's stargazer history stays visible in the README.


janosmiko (approved):

janosmiko:
Thanks @Dessalines39394 , I appreciate it.

Issue #646, opened 2026-08-20
[bug]: 19.8% CPU usage


fallenleavesgocrunch:
### Describe the bug

There's a lot of CPU usage going on when the program is parked on a page doing very little at all.

### Steps to reproduce

Get some details on the cluster info page - pin some statuses there - and it should happen.

### Expected behavior

Very very low. It's a TUI, it shouldn't be getting listed as excessive energy usage on my mac.

### Actual behavior

It's getting listed as using lots of energy :/

### lfk version

0.17.3

### Installation source

Homebrew

### Operating system

macOS

### Terminal emulator

WezTerm

### Relevant config

```yaml

```

### Relevant log output

```shell

```

### Pre-flight checks

- [x] I searched existing issues and did not find a duplicate.
- [x] I redacted any secrets, tokens, or personal data from this report.

fallenleavesgocrunch:
Success.

PR #656, opened 2026-08-20
perf: skip preview refetch when the hovered object is unchanged

janosmiko:
Part of the idle-CPU fix for #646.

- A background watch tick now fingerprints the hovered object's `resourceVersion` (scoped by context and kind) and skips the containers and YAML sub-fetches when it is unchanged.
- Metrics, events, secret data, and endpoints keep polling. They live in other API objects and never bump the pod RV.
- Manual refresh (`shift+r`), cursor movement, and drill-in always fetch. The skip only arms on `suppressBgtasks` paths.
- The fingerprint map self-bounds to one entry, so tab clones stay cheap.

## Test plan
- [x] `TestPreviewSkip_SameRVOnWatchTick` - no containers/YAML commands on an unchanged tick
- [x] `TestPreviewSkip_DifferentContextNeverSkips` - cross-context collision guard
- [x] `TestPreviewSkip_FingerprintStaysBounded` - map holds one entry
- [x] Manual refresh / bumped RV cases dispatch the full cascade
- [x] `go build`, `go vet`, `golangci-lint` (0 issues), `go test ./...` all pass


janosmiko:
@coderabbitai review

janosmiko:
@coderabbitai review

PR #661, opened 2026-08-20
perf: throttle list metrics refresh to a minimum interval

janosmiko:
Last piece of the idle-CPU fix for #646 (with #656 and #659). Pod and node list metrics fetched on every 2s watch tick although metrics.k8s.io only refreshes about every 15s.

- New `metrics_interval` config key: minimum interval between metrics list fetches. Default `15s`, clamp `[2s, 10m]`, `0` disables the throttle. Wired through the full config checklist (schema, reference, example, wiring tests).
- The gate only arms on background ticks. First view, manual refresh (`shift+r`), and drill-in always fetch.
- Keyed by context and kind, so clusters never share a cadence stamp.
- Node uptime stays unthrottled - it comes from Prometheus, not metrics-server.

## Test plan
- [x] Red-first tests: silent tick inside the interval skips, manual refresh inside the interval fetches, interval 0 fetches every tick, first load fetches immediately, Node path covered
- [x] Mutation check: neutering the guard makes the dispatch-site test fail
- [x] Config parse/clamp unit tests (7 cases) + all four config-sync enforcement tests pass
- [x] Go review and performance review: APPROVE
- [x] `go build`, `go vet`, `golangci-lint` (0 issues), `go test ./...` 26 packages, `-race` on internal/app


janosmiko:
@coderabbitai review

Issue #524, opened 2026-07-10
[bug]: Scrolling up breaks past the list you're looking at


fallenleavesgocrunch:
### Describe the bug

Much like a previous bug where scrolling down would escape the list you're scrolling and scroll parent lists - the same happens when you're scrolling up.

### Steps to reproduce

Go to any long sub list where the parent list isn't on item #1 and scroll up

### Expected behavior

Scroll up and stop scrolling at the top of the list you're scrolling

### Actual behavior

scrolls to the top of the list you're scrolling and then scrolls up on the parent list

### lfk version

0.15.1

### Installation source

Homebrew

### Operating system

macOS

### Terminal emulator

WezTerm

### Relevant config

```yaml

```

### Relevant log output

```shell

```

### Pre-flight checks

- [x] I searched existing issues and did not find a duplicate.
- [x] I redacted any secrets, tokens, or personal data from this report.

janosmiko:
Thanks — I tried hard to reproduce this one and couldn't trigger it so far: on both 0.15.1 and 0.15.5 I ran wheel and keyboard scrolling (ctrl+u/ctrl+b/pgup/shift+up/k) over all three panes at the contexts, resource-kinds, resource, and owned-pods levels, with the parent pane scrolled and with live content churn, and every list stops cleanly at the top.

Since this is the same clobber class as #398 (an internal render pass resetting the neighbouring pane's persistent scroll), I've put up #526 which enforces the #399 guard inside the renderers themselves — after that, no render path (known or not) can move the pane to the left anymore. It'll be in the next release.

To make sure that actually covers your case, could you share:

- which screen you're on when it happens (e.g. resource kinds list, a pod list, a Deployment's pods) and which pane the pointer is over,
- whether you scroll with a mouse wheel, trackpad, or keyboard,
- whether lfk runs inside tmux/screen/zellij within WezTerm,
- your lfk config if you have one (especially `scrolloff`, `mouse`, or keybinding overrides),
- ideally a short screen recording.

If you can still trigger it on the release after #526 lands, that info will let me nail the exact path.

fallenleavesgocrunch:
This isn't happening for me anymore so it must have been fixed.

fallenleavesgocrunch:
Services > a service - try scrolling on the details pane. That's where it goes crazy for me.

fallenleavesgocrunch:
[...]

PR #663, opened 2026-08-21
refactor: split the files and functions sitting at the lint caps

janosmiko:
Six files sat one or two lines under the 800-line revive cap and thirteen functions sat exactly at the gocyclo cap of 30. The next small edit in any of them failed CI mid-feature.

A `gocyclo -over 25` sweep corrected the picture before any code moved. `internal/ui/styles.go` was the one file actually at 800, not `app.go` at 799. Thirteen functions were bare at the cap, not nine, and ten more above 30 were already suppressed with `//nolint:gocyclo`, which is why nobody noticed. No suppression was added here.

## Files split

| File | Before | After | New file |
|---|---|---|---|
| `internal/ui/styles.go` | 800 | 491 | `styles_status.go`, `styles_condition.go` |
| `internal/app/view_overlays.go` | 799 | 523 | `view_overlays_fullscreen.go` |
| `internal/app/view_right.go` | 799 | 575 | `view_right_levels.go` |
| `internal/app/overlay_hintbar.go` | 798 | 362 | `overlay_hintbar_overlays.go` |
| `internal/app/update_resources_loaded.go` | 795 | 561 | `update_resources_discovery.go` |
| `internal/app/cursor.go` | 794 | 559 | `cursor_filter.go` |

## Functions off the cap

`computeVisibleMiddleItems` 30 to 5, `updateAPIResourceDiscovery` 30 to 20, `renderOverlayContentExtended` 30 to 20, `handleColorschemeNormalMode` 30 to 15, `handleLogTopKey` 30 to 17, `copyYAMLToClipboard` 30 to 18, `handleObjectExplorerNavKey` 30 to 19, `updateActionResultMsg` 30 to 21, `viewYAML` 30 to 24, plus `updateEditorResultMsg`, `handleMouse`, `updateEasterEggMsg` and `handleOverlayKeySecondary` under 25.

One behavior-adjacent change: `handleObjectExplorerWheel` lost its `tea.Cmd` return. Both branches always returned nil, matching the keyboard equivalents in `handleObjectExplorerNavKeyExtended`, and the object explorer holds its tree and preview in memory so there is no refetch to schedule. `unparam` only surfaced it once `handleModeWheel` was extracted.

`internal/app/app.go` stays at 799 and is not touched. It is a single `Model` struct, and Go cannot split a struct across files. The embedded sub-struct route forces rewrites of keyed `Model{...}` literals across 199 sites in 13 test files, so it needs its own change.

## Test plan

- [x] `go test -race ./...` — 15842 tests, 0 failures across 26 packages
- [x] `golangci-lint run ./...` — 0 issues on the whole tree
- [x] `gocyclo -over 25 ./internal/` — every remaining function over 30 carries a `//nolint`
- [x] `BenchmarkVisibleMiddleItemsUnchanged` still hits the memo at 7-8 ns/op, 0 allocations
- [ ] Manual pass in `--demo`: overlays, right pane, YAML viewer, object explorer, wheel momentum over the right pane (#524)

Each split was reviewed as a diff-level move: case-arm counts were compared against the base for every extracted dispatcher, so no key or message handler was silently dropped.


janosmiko:
All three findings check out as accurate readings of the code, and all three describe code this PR only moved. Each one was verified by diffing the flagged function against `origin/main`.

| Finding | Verdict |
|---|---|
| `cursor_filter.go:110` fuzzy score | `sortItemsByFuzzyScore` is the verbatim inner block of `computeVisibleMiddleItems` at `origin/main:internal/app/cursor.go:624-644`. The `ui.FuzzyScore(item.Name, query)` line is unchanged. |
| `update_resources_discovery.go:199` cache key | `handleAPIResourceDiscoveryError` diffs byte-identical against `origin/main:internal/app/update_resources_loaded.go:271`. |
| `update_resources_discovery.go:254` clamping | `updateDiscoveryCacheLoaded` diffs byte-identical against `origin/main:internal/app/update_resources_loaded.go:318`. Not one token changed. |

So none of them blocks this PR, which is a mechanical split of files at the 800-line revive cap and functions at the gocyclo cap of 30.

Two of the three were worth real work, and both now have their own PR stacked on this one:

- #665 fixes the discovery-error fallback. The level named in the finding, `LevelResourceTypes`, turned out to be safe, because back-navigation clears `nav.ResourceType` and `syncExpandedGroup` clamps. The reachable case is one level deeper at `LevelResources`: drill into a resource type before discovery answers, and a discovery failure replaces the list and poisons the cache. Reproduced, fixed, and pinned with a test that fails against the pre-fix file.
- #666 fixes the fuzzy ranking, scoring each item on the field `applyTextFilter` actually accepted rather than on `Name` alone. Measured to add no heap allocation on the render path.

The `syncExpandedGroup` mutation nitpick is declined: pointer-receiver mutation is the established convention throughout `cursor.go`. The hint-bar triplication predates this split and is left alone.

PR #664, opened 2026-08-21
fix: cancel superseded update listeners in the port-forward and capture managers

janosmiko:
`restorePortForwards` armed one `waitForPortForwardUpdate` listener per restore batch. Every `portForwardUpdateMsg` re-armed a new listener with a new channel while `SetUpdateCallback` held only one slot, so each superseded listener blocked forever on a channel nothing would ever write to. One goroutine leaked per restored port-forward.

`CaptureManager` had the same pair of bugs, so this fixes the cause once rather than patching one surface.

## What changed

The manager now owns the listener channel instead of calling a callback. `SetUpdateListener(ch chan<- struct{})` returns a `superseded` channel that closes when a newer listener replaces the old one, and `notifyLocked` does a non-blocking send while the manager holds its own mutex.

That ordering is the fix. Previously the manager captured the callback under the lock and invoked it after unlocking, so a supersession could slip into the gap and the listener would exit while the update was still in flight. Now every send completes inside the same critical section that guards the state being announced, so a listener that observes the close is guaranteed the drain finds its update.

Two shared pieces came out of it: `internal/k8s/update_listener.go` for the handshake, embedded by both managers, and `internal/app/commands_update_wait.go` for the drain-on-superseded waiter both commands use. `waitForPortForwardSignal` is now a two-line wrapper, so the existing port-forward tests still exercise the real path.

The helper deliberately carries no mutex of its own. Correctness depends on the owner calling `setLocked` and `notifyLocked` under the mutex guarding the state being announced. That contract is documented on the type and both methods, and the `-Locked` suffix follows the usual Go convention. A review checked all 11 call sites across both managers, not a sample.

Also here: `StopAll` now notifies when an entry actually changed state, and `CaptureEntry.onUpdate` is gone, having been assigned and never read.

## On the regression test

The first version of the stress test only proved that a mutex excludes. A review reproduced the pre-fix shape at each capture notify site and found the test shipped green for one of the three. The test now shrinks the Starting-to-Running flip delay through an injectable `captureRunningFlipDelay` and churns 16 concurrent listeners, which brings that site from 0 of 30 runs caught to 23 of 30.

Measured over 30 `-race` runs per site with the pre-fix shape restored:

| Notify site | Caught |
|---|---|
| append, Starting | 15 / 30 |
| decoder done | 25 / 30 |
| flip to Running | 23 / 30 |
| correct code | 0 / 30, no false positive |

This is a real scheduling race, so none of the three is deterministic, and the rates vary by machine. With `-count=5` in CI the weakest site misses roughly 3% of the time, against 100% before.

## Test plan

- [x] `go test -race ./...` — 23 packages ok. One unrelated pre-existing flake, `TestExplainFetchesStopWhenTheViewCloses`, passed 5 of 5 in isolation
- [x] `go test -race -count=3 ./internal/k8s/... ./internal/app/...` — no leaked goroutine, no timeout
- [ ] Manual pass against a live cluster: start three port-forwards, quit and relaunch so the session restores them, stop one and confirm the others keep updating; start a capture, stop it, restart it

### DIFF
--- a/internal/app/commands_capture.go
+++ b/internal/app/commands_capture.go
@@ -85,23 +85,19 @@ type kubesharkLaunchedMsg struct {
 // pattern.
 type captureUpdateMsg struct{}
 
-// waitForCaptureUpdate registers a manager-level update callback and returns a
-// tea.Cmd that blocks until the next state change. The handler is responsible
-// for re-arming via another waitForCaptureUpdate() — same pattern as
+// waitForCaptureUpdate blocks until the next capture state change. The handler
+// must re-arm it with another waitForCaptureUpdate(), same as
 // waitForPortForwardUpdate.
 func (m Model) waitForCaptureUpdate() tea.Cmd {
 	if m.captureMgr == nil {
 		return nil
 	}
 	ch := make(chan struct{}, 1)
-	m.captureMgr.SetUpdateCallback(func() {
-		select {
-		case ch <- struct{}{}:
-		default:
-		}
-	})
+	superseded := m.captureMgr.SetUpdateListener(ch)
 	return func() tea.Msg {
-		<-ch
+		if !waitForUpdateSignal(ch, nil, superseded) {
+			return nil
+		}
 		return captureUpdateMsg{}
 	}
 }
--- a/internal/app/commands_capture_test.go
+++ b/internal/app/commands_capture_test.go
@@ -68,7 +68,7 @@ func TestLaunchKubeshark_ReturnsTeaCmd(t *testing.T) {
 	// Don't call cmd() — would try to actually port-forward + open browser.
 }
 
-func TestWaitForCaptureUpdate_DispatchesMsgWhenCallbackFires(t *testing.T) {
+func TestWaitForCaptureUpdate_StaysBlockedWithoutAStateChange(t *testing.T) {
 	m := baseFinalModel()
 	m.captureMgr = k8s.NewCaptureManager()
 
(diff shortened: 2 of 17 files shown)


janosmiko:
Fixed, and the finding reached further than the one line.

`logger.Error` does not redact centrally, so the finding was real. `logger.Redact` and `logger.RedactErr` already exist in `internal/logger/redact.go` from earlier work, so the fix routes the value through the redactor rather than dropping the field. The error text is what a user needs when a port-forward fails, so deleting it would trade a security issue for a debuggability one.

Two changes beyond the suggested patch:

**A second site.** `internal/k8s/capture.go:351` stores raw `kubectl` stderr in `CaptureEntry.LastError`, which `overlay_traffic_capture.go` renders on screen. Same symptom, so it got the same fix rather than a second patch later.

**Redaction happens at the assignment, not at the log call.** That covers the log line, the status bar and the overlay in one place, instead of leaving the raw value in the struct for the next reader to leak.

**An ordering bug the finding did not reach.** In `capture.go` the stderr was trimmed to a 240-character tail before being stored. Trimming first can cut the `ya29.` prefix off an OAuth token, leaving a tail that no redaction pattern matches, which then gets stored raw. Redaction now runs before the trim. `TestClassifyCaptureExit_RedactsBeforeTrimmingLongStderr` pins that order, and it leaked a 36-character raw fragment when checked against the trim-first shape.

Four tests in `internal/k8s/subprocess_stderr_redaction_test.go`, each red-checked by reverting its own edit. The port-forward one reproduced a raw `ya29…` token in the log output.

Checked and left alone: `describe_pod.go:32` already uses `RedactErr`, and `translateKubectlDebugErr` returns only fixed strings.

janosmiko:
@coderabbitai review

janosmiko:
Declining the in-place mutation finding on this PR, with the evidence, and filing the one real invariant it surfaced.

The guideline is right in general. Here it changes nothing observable, because entries never escape the mutex as live pointers.

**All four flagged sites hold `m.mu`.** `capture.go:303-310` (`markCaptureRunning`, Lock plus defer at 304), `portforward.go:307-316` (stdout monitor, explicit Lock/Unlock around both writes), `portforward.go:331-353` (`recordProcessExit`, Lock plus defer at 332), `portforward.go:424-434` (`StopAll`, Lock plus defer at 421).

**Every accessor returns value copies, not live pointers.** `Entries` at `portforward.go:104` does `result[i] = *e`, `EntriesForDisplay` at `:117` does `entry := *e`, `GetEntry` at `:407` returns a pointer to a copy, and `capture.go:117` copies field by field with the atomics loaded.

**Goroutines do retain entry pointers** (`portforward.go:289`, `capture.go:263`, `capture.go:289`), but every field write re-acquires `m.mu`. The only lock-free writes are `PacketCount` and `ByteCount`, which go through `atomic.AddInt64` in `packet_decoder.go:119-120` and are read with `atomic.LoadInt64` at `capture.go:128-129`.

**The TUI never holds a live entry across a frame.** It consumes value slices and converts immediately into its own structs at `tabs_portforward.go:16` and `captures.go:202-205`, and `commands_portforward.go:78` uses a `GetEntry` copy.

So no reader can observe a torn entry, and the rewrite this asks for is an architectural change to the manager's state model on a PR scoped to listener cancellation. The mutation pattern also predates this branch.

The review did surface one thing worth keeping: `CaptureEntry` must never be plain-copied outside `Entries()`, because of those two atomic counters. That is a real invariant with no guard on it today, and it is tracked separately.

PR #665, opened 2026-08-21
fix: keep the discovery-error fallback from clobbering a deeper level

janosmiko:
Stacked on #663, which is where `internal/app/update_resources_discovery.go` lives. Merge that one first.

## The bug

Enter a context, then drill into a resource type before the apiserver answers the discovery request. If discovery then fails, `handleAPIResourceDiscoveryError` replaced your pod list with the seed resource-type sidebar and wrote those resource types into `itemCache["<ctx>/pods"]`, so going back and re-entering served the poisoned entry.

The fallback guarded only on `m.loading` and keyed the cache with `m.navKey()`, neither of which says anything about which level the user is on.

Entering a context fires `discoverAPIResources` (`update_navigation.go:409`), so the window is real: the user drills in while that request is still outstanding.

## The fix

The fallback now guards `isCurrentContext && m.nav.Level == model.LevelResourceTypes && m.loading`, mirroring the success branch, and keys the cache with `m.nav.Context`.

The key change is a no-op in every state the new guard can reach, since `nav.ResourceType`, `ResourceName` and `OwnedName` are all cleared at that level. It is there to make the intent explicit rather than to fix a second bug.

One asymmetry against the success branch is deliberate. The success path caches unconditionally when the context matches, even at deeper levels, because it has real discovered data. The error path caches only inside the level guard, because on failure it has nothing but a seed list, and writing that while the user is deeper is the bug.

## Where CodeRabbit had it

CodeRabbit flagged this on #663 and named `LevelResourceTypes` as the vulnerable level. That level is actually safe: back-navigation clears `nav.ResourceType`, so `navKey()` already equals `nav.Context`, and `syncExpandedGroup` clamps the cursor. The reachable case is one level deeper, at `LevelResources`. The finding was worth chasing, but not where it pointed.

## Test plan

- [x] `TestDiscoveryFailureAtResourceLevelLeavesMiddleColumnAlone` — fails against the pre-fix file, passes now. Verified by running it against a copy of the parent's `update_resources_discovery.go`
- [x] `TestDiscoveryFailureAtResourceTypesLevelSeedsAndCachesUnderContext` — pins the intentional fallback so a future change cannot quietly remove it
- [x] `go test -race ./...` — 26 packages ok. `TestExplainFetchesStopWhenTheViewCloses` is a known unrelated flake and passes in isolation
- [ ] Manual pass: point a context at an unroutable server, drill into a resource type before discovery answers, confirm the list is not replaced when it fails

Every site that sets `m.loading` before firing discovery sets the level to `LevelResourceTypes` first, so the tighter guard drops no legitimate case and cannot leave a spinner stuck.

### DIFF
--- a/internal/app/discovery_failure_level_test.go
+++ b/internal/app/discovery_failure_level_test.go
@@ -1,0 +1,73 @@
+package app
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/janosmiko/lfk/internal/model"
+)
+
+// Entering a context fires discovery, and the user can drill into a resource
+// type before the apiserver answers. A failure arriving then must not paint the
+// seed resource types over the resource list, nor cache them as "test-ctx/pods".
+func TestDiscoveryFailureAtResourceLevelLeavesMiddleColumnAlone(t *testing.T) {
+	t.Parallel()
+	m := baseModelCov()
+	m.nav.Level = model.LevelResources
+	m.nav.Context = "test-ctx"
+	m.nav.ResourceType = model.ResourceTypeEntry{Kind: "Pod", Resource: "pods"}
+	m.allGroupsExpanded = true
+
+	pods := []model.Item{
+		{Name: "api-0", Kind: "Pod"},
+		{Name: "api-1", Kind: "Pod"},
+		{Name: "api-2", Kind: "Pod"},
+	}
+	m.setMiddleItems(pods)
+	m.setCursor(2)
+	// The pod list is still being fetched, so the loader is up.
+	m.loading = true
+
+	updated, _ := m.updateAPIResourceDiscovery(apiResourceDiscoveryMsg{
+		context: "test-ctx",
+		err:     errors.New("the server is currently unable to handle the request"),
+	})
+
+	assert.Equal(t, pods, updated.middleItems,
+		"a discovery failure while the user is at LevelResources must not "+
+			"replace the resource list with the resource-type seed list")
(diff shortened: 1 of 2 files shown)


janosmiko:
@coderabbitai review

janosmiko:
Declining the `itemCache` copy, with the mechanism that makes the write safe.

The reading is correct that `Model` has a value receiver and line 202 writes into the map in place. It is safe because isolation is enforced once, at the tab boundary, not at each write site.

`copyItemCache` (`internal/app/tabs_copy_helpers.go:40`) deep-copies the map: new map, new slices, and cloned `Columns`, `Conditions` and `GroupedRefs` per item. It runs at all three seams, `saveCurrentTab` (`tab_state.go:83`), `loadTab` (`tab_state.go:226`) and `cloneCurrentTab` (`tab_state.go:479`). Those are the only field-level assignments outside tests, apart from `update_bookmarks.go:702`, which replaces the map wholesale. No `TabState` and no live `Model` ever hold the same map header, so a write through it cannot reach another tab. Cache reads in `commands_load.go` extract the slice before the closure, so no goroutine touches the map either.

That inner-slice cloning inside `copyItemCache` is the scar from a real past bug in the `Item.Columns` enrichment path, which had exactly this shape. It was fixed at the clone boundary, which is why per-site copies are not needed now.

Roughly 20 sites across 12 files write `m.itemCache[k] = …` in place, including the success branch of this very message eight lines above the flagged line, writing the same `m.nav.Context` key. Copying here would allocate a fresh map on every discovery update, on a per-navigation path, and leave one copying site among twenty in-place writers. That is worse than consistent in-place writes, and it would hide where the invariant actually lives.

No other findings on this PR.

Issue #398, opened 2026-06-10
[bug]: scroll wheel scrolls left list when current list is exhausted


fallenleavesgocrunch:
### Describe the bug

Mouse wheel scrolling what's underneath the mouse works, except when you get to the end of the content - then in some situations (such as the main resource kinds list) it will suddenly scroll the lefter tab.

### Steps to reproduce

Pane 1: 2+ items.
Pane 2: More than a screens height of items.
Pane 3: More than screens height of data.

Mouse in Pane 2 (definitely happens) or Pane 3 (sometimes happens) and wheel scroll.

### Expected behavior

Get to the bottom of the list and stop scrolling.

### Actual behavior

Gets to the bottom of the list, and then scrolls down on the pane to the left (ie: which cluster you're looking at).

### lfk version

0.14.1

### Installation source

Homebrew

### Operating system

macOS 26.5

### Terminal emulator

WezTerm

### Relevant config

```yaml

```

### Relevant log output

```shell

```

### Pre-flight checks

- [x] I searched existing issues and did not find a duplicate.
- [x] I redacted any secrets, tokens, or personal data from this report.

janosmiko:
Thanks for the report — reproduced and root-caused, fix is up in #399.

What you were seeing: wheel events over a pane were routed correctly, but every wheel tick over the **preview pane** (and, transitively, content refreshes while scrolling) ran an internal "clamp scroll to content height" step that *renders* right-pane content to measure it. That render path unintentionally wrote to the persistent scroll state of the **middle/left pane**, resetting it to 0. The next frame re-derived that pane's viewport from scratch — which is why the pane to the left of the one you were scrolling suddenly jumped, and why it showed up most reliably once you hit the end of a list (that's when your viewport position differs most from the re-derived one).

Two flavors, matching what you observed:
- at the resource list level with a split preview (e.g. Deployments/Pods), the clobber fired on **every** wheel tick — "definitely happens"
- elsewhere it fired only when the preview content changed (loads, watch refreshes) — "sometimes happens"

The fix guards that measurement path the same way the normal render path already does, so wheel scrolling now stops cleanly at the end of a list. It'll be in the next release.

fallenleavesgocrunch:
I was about to add that it also happened on scroll up but you've already fix it. Nice work.

PR #667, opened 2026-08-21
fix: keep the discovery-error fallback from clobbering a deeper level

janosmiko:
## The bug

Enter a context, then drill into a resource type before the apiserver answers the discovery request. If discovery then fails, `handleAPIResourceDiscoveryError` replaced your pod list with the seed resource-type sidebar and wrote those resource types into `itemCache["<ctx>/pods"]`, so going back and re-entering served the poisoned entry.

The fallback guarded only on `m.loading` and keyed the cache with `m.navKey()`, neither of which says anything about which level the user is on.

Entering a context fires `discoverAPIResources` (`update_navigation.go:409`), so the window is real: the user drills in while that request is still outstanding.

## The fix

The fallback now guards `isCurrentContext && m.nav.Level == model.LevelResourceTypes && m.loading`, mirroring the success branch, and keys the cache with `m.nav.Context`.

The key change is a no-op in every state the new guard can reach, since `nav.ResourceType`, `ResourceName` and `OwnedName` are all cleared at that level. It is there to make the intent explicit rather than to fix a second bug.

One asymmetry against the success branch is deliberate. The success path caches unconditionally when the context matches, even at deeper levels, because it has real discovered data. The error path caches only inside the level guard, because on failure it has nothing but a seed list, and writing that while the user is deeper is the bug.

## Where CodeRabbit had it

CodeRabbit flagged this on #663 and named `LevelResourceTypes` as the vulnerable level. That level is actually safe: back-navigation clears `nav.ResourceType`, so `navKey()` already equals `nav.Context`, and `syncExpandedGroup` clamps the cursor. The reachable case is one level deeper, at `LevelResources`. The finding was worth chasing, but not where it pointed.

## Test plan

- [x] `TestDiscoveryFailureAtResourceLevelLeavesMiddleColumnAlone` — fails against the pre-fix file, passes now. Verified by running it against a copy of the parent's `update_resources_discovery.go`
- [x] `TestDiscoveryFailureAtResourceTypesLevelSeedsAndCachesUnderContext` — pins the intentional fallback so a future change cannot quietly remove it
- [x] `go test -race ./...` — 26 packages ok. `TestExplainFetchesStopWhenTheViewCloses` is a known unrelated flake and passes in isolation
- [ ] Manual pass: point a context at an unroutable server, drill into a resource type before discovery answers, confirm the list is not replaced when it fails

Every site that sets `m.loading` before firing discovery sets the level to `LevelResourceTypes` first, so the tighter guard drops no legitimate case and cannot leave a spinner stuck.

Supersedes #665, which GitHub closed when its base branch was deleted on merge. The CodeRabbit review there raised one finding about copying `itemCache`, declined with the mechanism in that thread: `copyItemCache` deep-copies at all three tab seams, so an in-place write cannot leak across tabs.

### DIFF
--- a/internal/app/discovery_failure_level_test.go
+++ b/internal/app/discovery_failure_level_test.go
@@ -14,8 +14,10 @@ import (
 // type before the apiserver answers. A failure arriving then must not paint the
 // seed resource types over the resource list, nor cache them as "test-ctx/pods".
 func TestDiscoveryFailureAtResourceLevelLeavesMiddleColumnAlone(t *testing.T) {
-	t.Parallel()
-	m := baseModelCov()
+	// Not t.Parallel(): View() writes ui.ActiveCollapsedCategories, a
+	// package-level render global, so two Model.View() calls racing across
+	// goroutines trip -race.
+	m := basePush80Model()
 	m.nav.Level = model.LevelResources
 	m.nav.Context = "test-ctx"
 	m.nav.ResourceType = model.ResourceTypeEntry{Kind: "Pod", Resource: "pods"}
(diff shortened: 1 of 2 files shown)


janosmiko:
Done, and it surfaced a real data race on the way.

Both tests now build with `basePush80Model()` and assert on `stripANSI(View().Content)`, on top of the existing state checks rather than replacing them. The state assertions cover the half a user cannot see, namely the poisoned cache entry.

- `TestDiscoveryFailureAtResourceLevelLeavesMiddleColumnAlone` asserts `api-0`, `api-1` and `api-2` are still rendered and that `Deployments` does NOT appear in the middle column.
- `TestDiscoveryFailureAtResourceTypesLevelSeedsAndCachesUnderContext` asserts `Deployments` and `ConfigMaps` do appear, since that fallback is intentional.

The view assertions were mutation-checked rather than assumed: reverting `handleAPIResourceDiscoveryError` to the pre-fix guard and cache key in place made them fail exactly as expected, with the sidebar painted into the middle column. The file was then restored byte for byte, confirmed with `git diff`.

**Unplanned finding.** Adding a second `View()`-rendering test under `t.Parallel()` failed under `-race`, and it is a genuine production race, not a test artifact: `Model.View()` writes the package-level `ui.ActiveCollapsedCategories` and `ui.ActiveCategoryCounts`, so two `View()` calls from different `Model` instances on different goroutines collide. The trace pins it at a write in `view.go:439` against a read in `explorer.go:351`, colliding with the existing `TestWatchTickAgeAdvancesOnMemoizedItems`, which also renders under `t.Parallel()`.

Bubble Tea renders on one goroutine, so this is latent rather than live. Both tests dropped `t.Parallel()` with a comment naming the reason, which keeps this PR in scope. The underlying globals are tracked separately, in the same family as the cursor-less render invariant from #526 and #398.

janosmiko:
@coderabbitai review

PR #688, opened 2026-08-24
fix: respect the kubeconfig context namespace at startup

### LINKED ISSUE #686: [feat]: namespace to be available in the config
### Problem / motivation

Would be great if we can specify the namespace in the lfk's config file or lfk to respect the namespace set in the KUBECONFIG file.

### Proposed solution

`namespace` parameter in the lfk's config file (e.g. `namespace: kube-system`) OR lfk to respect the namespace set in KUBECONFIG file.

### Alternatives considered

_No response_

### Area

Configuration

### Additional context

_No response_

### Pre-flight checks

janosmiko:
Closes #686

`lfk` already read `contexts[*].namespace` from the kubeconfig into the model, but the `all_namespaces` startup default of `true` masked it. A context that pinned a namespace still opened on every namespace, so the setting looked ignored.

## Precedence

Highest first:

1. `--namespace` flag
2. restored session tab namespace
3. `all_namespaces` set explicitly in the config file (true = all, false = context namespace)
4. kubeconfig `contexts[*].namespace`, when the context pins one
5. all-namespaces default

No new config key. The kubeconfig is the per-context source of truth, which is what a multi-cluster tool needs.

## Changes

- `resolveStartupAllNamespaces` decides the startup scope. An explicit `all_namespaces` wins. Otherwise a pinned context namespace scopes the view.
- `rescopeNamespaceForContext` re-applies the same rule on a plain context switch, so each cluster opens where its own kubeconfig says. Wired at two sites only: entering a cluster from the clusters list, and `:ctx <name>`. Bookmarks, sessions, and union sets carry their own namespace and stay untouched.
- The re-scope drops the namespace selection state that belongs to the cluster you left, including the all-namespaces stash. Without that, a later `A`-toggle or previous-namespace jump restored a namespace name from the previous cluster. `namespace_history.go` already drops the same pair for the same reason.
- `lfk --context foo` with no `-n` no longer forces all-namespaces mode. It honoured neither the config setting nor the pinned namespace before.

## Test plan

- [x] `go build ./...`
- [x] `go test -count=1 ./internal/app/... ./internal/ui/... ./internal/k8s/...`
- [x] 9 new tests in `internal/app/namespace_context_scope_test.go`, one per precedence rule plus the stash-leak guard
- [x] Mutation-checked the stash-leak guard: it fails when the clearing lines are removed, passes when restored
- [ ] Manual check against a kubeconfig whose context pins a namespace

### DIFF
--- a/docs/config-example.yaml
+++ b/docs/config-example.yaml
@@ -234,10 +234,9 @@ log_path: /tmp/lfk.log
 #   tree: false  # open in the ASCII-art tree view (toggle: T)
 
 # ─── Session Defaults ────────────────────────────────────────────────────────
-# Startup state for the main view. split_preview toggles the preview pane;
-# watch_mode toggles live polling; all_namespaces sets the namespace scope
-# (false starts scoped to the context's default namespace). The --namespace CLI
-# flag and per-bookmark/session scope override all_namespaces.
+# Startup state for the main view. split_preview toggles the preview pane,
+# watch_mode toggles live polling, all_namespaces sets the namespace scope
+# (unset, a pinned kubeconfig namespace wins). CLI/session scope overrides it.
 # split_preview: true
 # watch_mode: true
 # all_namespaces: true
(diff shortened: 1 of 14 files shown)


janosmiko:
@coderabbitai review

PR #700, opened 2026-08-26
fix: use interactive bash when available

sysjonny:
## What

Improve the shell used when executing into Linux containers.

## Changes

- Check whether `clear` is available before executing it.
- Start `bash` with `-i` when available.
- Keep `ash` as the fallback shell.
- Keep `sh` as the final fallback.

## Motivation

Some minimal containers do not provide the `clear` command, which currently
causes an error when opening a shell.

Additionally, using `bash -i` provides a proper interactive Bash session when
Bash is available.

## Fallback order

1. `bash -i`
2. `ash`
3. `sh`

`clear` is executed only when available.


janosmiko (commented):
Well, the automated tests are missing and it worth to check coderabbit's comments too.

janosmiko (approved):

Issue #705, opened 2026-08-28
[feat]: Document right-sizing advisor usage guidelines


macrokernel:
### Problem / motivation

I wonder if right-sizing advisor is practically useful? Please see 2 advisories generated for the same pod within a few minutes. They are very contradictory.

<img width="2494" height="450" alt="Image" src="https://github.com/user-attachments/assets/6a327e68-d722-44ce-8520-471b1440bd9b" />

<img width="2496" height="437" alt="Image" src="https://github.com/user-attachments/assets/5d8e5d21-b915-4c94-8372-1bda25caf429" />

### Proposed solution

Please write a short usage guide on how to get a useful sizing advisory.

### Alternatives considered

Perhaps the advisor should inspect pod resource usage on a larger time span like a few hours or days?

### Area

Action menu

### Pre-flight checks

- [x] I searched existing issues and did not find a duplicate.

janosmiko:
You are using the `snapshot` strategy, which fetches the current usage and it shows suggestion based on that +25% headroom. 
- On the first screenshot it uses 670m CPU, it suggest to raise the requests/limits
- On the second screenshot your pod only uses 70m CPU so it shows to to lower the requests/limits.

You can change the strategy from `snapshot` to `1d-peak`, `1d-avg`, or `7d-p95` by pressing `[` or `]` hotkeys.
You can also change the headroom by pressing `<` or `>`.

These are shown on the hint bar on the bottom.

I hope this helps.

<img width="2514" height="2046" alt="Image" src="https://github.com/user-attachments/assets/9394f6a8-49eb-4ad3-af79-b1b7fd2b994a" />

macrokernel:
Thanks for the clarification - it's very nice to have this in the docs. Lfk is not showing `[/]` strategy selection hint on my clusters. Is it because Prom metrics are unavailable?

<img width="2963" height="1342" alt="Image" src="https://github.com/user-attachments/assets/60cde9d5-b61f-41cb-bb73-826e19084ae1" />

janosmiko:
Yes, that's the reason for that. It works based on VPA or Prometheus, but if no metrics available, it falls back to metrics-server and shows the current usage snapshot and in your case only that's available.

macrokernel:
Hi @janosmiko,

I still cannot figure out how to use right-sizing feature. Can you please help?

[...]

PR #717, opened 2026-09-01
fix: keep the right-sizing window across a workload rollout

janosmiko:
Closes the bug reported in #705 (comment of 2026-08-31): applying a `7d-p95` recommendation and re-checking in the same mode produced a second recommendation about 40 percent different.

## Cause

`buildPromContainerQuery` built its selector from `resolvePodsForWorkload`, which returns the pods running at query time:

```promql
pod=~"^(frontend-5x9dkm2t7b-q4wzn|frontend-5x9dkm2t7b-j2vbn)$"
```

Applying new requests or limits changes the pod spec, so the Deployment rolls and every pod name changes. The next query matched only pods that were minutes old, and `quantile_over_time(0.95, ...[7d:5m])` returned the p95 of those minutes. The overlay header still printed `window: 7d`, so nothing showed that the window was empty.

This also hits any workload that restarted inside the window, not only one that was just resized, and it applies to `1d-peak` and `1d-avg` as well.

## Fix

Select pods by the name shape the controller produces, so the window also covers revisions that no longer exist:

```promql
pod=~`^frontend-[bcdfghjklmnpqrstvwxz2456789]{1,10}-[bcdfghjklmnpqrstvwxz2456789]{5}$`
```

`podsRegexForWorkload` has a pattern per kind, following each controller's naming rule: Deployment (`<name>-<pod-template-hash>-<suffix>`), StatefulSet (ordinal), DaemonSet, Job (with the optional Indexed completion index), CronJob (Job named after its scheduled minute), and bare Pod. An unsupported kind returns `""` and the caller skips the query rather than selecting the whole namespace.

The character class is `rand.alphanums` from `k8s.io/apimachinery/pkg/util/rand`, the alphabet k8s generates name segments from. It has no vowels, which is what keeps a `foo` pattern off the pods of a sibling `foo-api`.

The pod matcher also moves into a raw PromQL literal (backticks). A double-quoted literal processes escapes and rejects the `\.` that `regexp.QuoteMeta` emits for a dotted workload name.

## Trade-offs

- The query now also reads pods from earlier revisions, which is where the history lives. A workload that was OOMKilled at its old limit will have its memory p95 capped by that limit, so the recommendation reads low. That was already true of any window longer than the current pods' age.
- Matching by shape is a heuristic, not a guarantee. A sibling workload whose extra name segment is all consonants and digits could match. Requires an unusual name and the same namespace.
- A workload name long enough for k8s to truncate the ReplicaSet name will not match its own pods. Noted in the code.

## Not in this PR

The overlay header still reports the selected window rather than the span the data actually covers. Filed separately.

## Test plan

- [x] `podsRegexForWorkload` table test: match and reject cases for all six kinds, including an Indexed Job and an Indexed CronJob
- [x] `TestPodsRegexForWorkload_SurvivesRollout`: pods from two different pod-template-hashes both match
- [x] `TestBuildPromContainerQuery_NotPinnedToCurrentPodNames`: extracts the matcher from the built query and checks both revisions against it
- [x] Every new test verified red against the old pinned-pod behaviour, not just green against the new code
- [x] `go test ./...` green, `make lint` reports 0 issues

### DIFF
--- a/internal/k8s/client_rightsizing.go
+++ b/internal/k8s/client_rightsizing.go
@@ -161,7 +161,7 @@ func (c *Client) GetRightsizing(ctx context.Context, contextName, namespace, kin
 	case model.StrategyVPA:
 		c.applyVPAStrategy(ctx, contextName, namespace, kind, name, headroom, out)
 	case model.StrategyPromMax1D, model.StrategyPromAvg1D, model.StrategyPromP957D:
-		c.applyPrometheusStrategy(ctx, contextName, namespace, pods, strategy, headroom, out)
+		c.applyPrometheusStrategy(ctx, contextName, namespace, kind, name, strategy, headroom, out)
 	case model.StrategySnapshot:
 		c.applySnapshotStrategy(ctx, contextName, namespace, pods, headroom, out)
 	}
--- a/internal/k8s/client_rightsizing_pods.go
+++ b/internal/k8s/client_rightsizing_pods.go
@@ -4,8 +4,38 @@ import (
 	"context"
 	"fmt"
 
+	batchv1 "k8s.io/api/batch/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+// workloadIsIndexed reports whether the workload's pods carry a
+// completion-index segment in their names. False on any lookup error,
+// which yields the narrower pod pattern.
+func (c *Client) workloadIsIndexed(ctx context.Context, contextName, namespace, kind, name string) bool {
+	cs, err := c.clientsetForContext(contextName)
+	if err != nil {
+		return false
+	}
+	switch kind {
+	case "Job":
+		job, err := cs.BatchV1().Jobs(namespace).Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return false
+		}
+		return isIndexedCompletion(job.Spec.CompletionMode)
+	case "CronJob":
+		cj, err := cs.BatchV1().CronJobs(namespace).Get(ctx, name, metav1.GetOptions{})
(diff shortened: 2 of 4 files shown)


janosmiko:
Confirmed and fixed in 05572904.

Reproduced both cases before changing anything:

| Pattern for | Pod of | Matched |
|---|---|---|
| Job `backup` | Job `backup-3` | yes |
| CronJob `nightly` | CronJob `nightly-2` | yes |
| Deployment `frontend` | Deployment `frontend-2` | no |
| StatefulSet `postgres` | StatefulSet `postgres-1` | no |

Only Job and CronJob were affected, because only they carried an optional `(?:[0-9]+-)?` group for an Indexed job's completion index. That group happily matched a sibling's numeric name segment instead.

I did not take the owner-metadata join. It needs kube-state-metrics, and the user who reported #705 runs VictoriaMetrics scraping cadvisor without it, so that route would trade this bug for no data at all on the setup the PR is meant to fix.

Instead the index group is now conditional: `workloadIsIndexed` reads `completionMode` from the Job, or from the CronJob's job template, and the group is added only when the workload really is Indexed. A failed lookup falls back to the narrower pattern, so the failure mode is missing data rather than a wrong number.

Rejection tests added for both sibling names, plus separate indexed and non-indexed table rows for Job and CronJob, and a test that `workloadIsIndexed` reads the spec. All four verified red against the always-on index group.

Worth noting the blast radius was narrower than "aggregates foreign workload metrics" suggests: the aggregation is `max by (container)`, and only containers whose names match one of ours are ever read out of the result map. A colliding sibling had to also share a container name to affect anything. Still wrong, still worth fixing.

PR #776, opened 2026-09-13
fix: offer prometheus rightsizing strategies when prometheus is discovered

janosmiko:
- Rightsizing offered the Prometheus strategies (`[`/`]`) only when `monitoring` config named Prometheus. Sparklines also use a Prometheus that label discovery finds, so users saw sparklines but no strategies.
- `AvailableRightsizingStrategies` now uses the same `prometheusAvailable` check as pod and node metrics.

Refs #705

## Test plan
- [x] Unit test: discovered Prometheus Service, no config, strategies listed
- [x] Unit test: no config, no Service, snapshot only
- [x] `go test -race ./...`, lint
- [ ] Manual: cluster with kube-prometheus-stack, no `monitoring` config, open rightsizing on a Deployment, `[`/`]` hint shows


fallenleavesgocrunch:
Fixed it.

